Issue #33, opened 2022-07-26
Remove express


DiegoRBaquero:
L1 doesn't need a full web framework. Express should be removed.

juliangruber:
I personally like using little wrappers like https://github.com/electron/update.electronjs.org/blob/f0d89fe197ed13b63fbe58af96492eab1b4c21cc/index.js#L42-L60, which help with error responses and request logging

juliangruber:
Instead of going frameworkless, @bajtos suggested looking at [fastify](https://ghub.io/fastify)

juliangruber:
Should we not remove `express`, we must secure the current `async` route handlers, which aren't supported natively by the framework. @guanzo suggested using https://www.npmjs.com/package/express-async-errors in this case.

PR #34, opened 2022-07-26
Harden container/shim/src/index.js

juliangruber:
I skipped the express specific parts for this review because of https://github.com/filecoin-saturn/L1-node/issues/33. Hope you find this useful!

### DIFF
--- a/container/shim/src/index.js
+++ b/container/shim/src/index.js
@@ -3,6 +3,8 @@ import fsPromises from 'node:fs/promises'
 import { cpus } from 'node:os'
 import express from 'express'
 import mimeTypes from 'mime-types'
+import timers from 'node:timers/promises'
+import assert from 'node:assert'
 
 import { addRegisterCheckRoute, deregister, register } from './modules/registration.js'
 import {
@@ -46,15 +48,17 @@ if (cluster.isPrimary) {
   process.on('SIGQUIT', shutdownCluster)
   process.on('SIGINT', shutdownCluster)
 
-  setTimeout(async function () {
-    await register(true).catch(err => {
-      debug(`Failed to register ${err.name} ${err.message}`)
-      process.exit(1)
-    })
+  await timers.setTimeout(500)
 
-    // Start log ingestor
-    await initLogIngestor()
-  }, 500)
+  try {
+    await register(true)
+  } catch (err) {
+    debug(`Failed to register ${err.name} ${err.stack}`)
+    process.exit(1)
+  }
+
+  // Start log ingestor
+  await initLogIngestor()
 } else {
   const ipfsAgent = new https.Agent({
     keepAlive: true,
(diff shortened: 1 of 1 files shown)


container/shim/src/index.js
```diff
@@ -46,15 +48,17 @@ if (cluster.isPrimary) {
   process.on('SIGQUIT', shutdownCluster)
   process.on('SIGINT', shutdownCluster)
 
-  setTimeout(async function () {
-    await register(true).catch(err => {
-      debug(`Failed to register ${err.name} ${err.message}`)
-      process.exit(1)
-    })
+  await timers.setTimeout(500)
```
juliangruber, line 51: safer to use the promise version

container/shim/src/index.js
```diff
@@ -46,15 +48,17 @@ if (cluster.isPrimary) {
...
+
+  try {
+    await register(true)
+  } catch (err) {
+    debug(`Failed to register ${err.name} ${err.stack}`)
+    process.exit(1)
+  }
```
juliangruber, line 58: To me it's easier to basic the standard language features

container/shim/src/index.js
```diff
@@ -91,9 +95,10 @@ if (cluster.isPrimary) {
     })
 
     if (req.headers.range) {
-      let [start, end] = req.headers.range.split('=')[1].split('-')
-      start = parseInt(start, 10)
-      end = parseInt(end, 10)
+      const range = req.headers.range.split('=')?[1]?.split('-')
```
juliangruber, line 98: an invalid range header could have crashed the process I think

container/shim/src/index.js
```diff
@@ -91,9 +95,10 @@ if (cluster.isPrimary) {
...
+      assert(range, 'TODO')
+      const start = Number(range[0])
+      const end = Number(range[0])
```
juliangruber, line 101: `Number` is safer than `parseInt` because you can't forget the radix

container/shim/src/index.js
```diff
@@ -120,6 +125,8 @@ if (cluster.isPrimary) {
       controller.abort()
     }, GATEWAY_TIMEOUT)
 
+    // Use node-fetch or undici instead, the `"error"`
+    // event makes this difficult to consume
```
juliangruber, line 129: I didn't know which way you wanted to go, but I think it is important to use a promise native request library.

container/shim/src/index.js
```diff
@@ -91,9 +95,10 @@ if (cluster.isPrimary) {
     })
 
     if (req.headers.range) {
-      let [start, end] = req.headers.range.split('=')[1].split('-')
-      start = parseInt(start, 10)
-      end = parseInt(end, 10)
+      const range = req.headers.range.split('=')?[1]?.split('-')
+      assert(range, 'TODO')
```
DiegoRBaquero, line 99: Do we want to fail on invalid range? Or return empty bytes?

juliangruber: I think a 4xx response is appropriate, so the client can know they need to adjust their request

DiegoRBaquero: Then we need either try catch or do a truthy check returning 400

DiegoRBaquero, line 99:
```suggestion
      if(!range) {
        return res.status(400).send('Invalid range')
      }
```

juliangruber: This is where the error handling comes in. I actually meant to use https://www.npmjs.com/package/http-assert:

```suggestion
      httpAssert(range, 400, 'invalid range')
```

If the async route handler now forwards this to `next`, then the express final middleware can take care of responding with the appropriate error per request

DiegoRBaquero: Can you add the update it with it?

PR #35, opened 2022-07-27
Add Brotli to NGINX

### LINKED ISSUE #21: add brotli compression support
brotli is widely supported in browsers (https://caniuse.com/brotli) and enough superior to gzip to use it over gzip when supported

see https://docs.nginx.com/nginx/admin-guide/dynamic-modules/brotli/ and http://wiki.centos-webpanel.com/enabling-brotli-compression-on-nginx

there has to be some way to add brotli support without nginx plus.. but if nginx plus keeps being an impediment to our various usage, ie death by a thousand cuts, perhaps its worth migrating to another http server in the future (😢)

enolal826:
Fixes #21.

This PR contains:
- A new Dockerfile that uses a multi-stage build to:
(Stage 1) Compile from source Nginx, get the Brotli and generate the `.so` files for the modules.
(Stage 2) By using the official `nginx` Docker image, the modules are moved to the corresponding directory.
- Changes in the `nginx.conf` file to enable Brotli, along with the supported types.

NOTE: `text/html` is always compressed, but I have added it to the supported types, just in case this default behavior changes in the future.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,69 @@
-# pull the Node.js Docker image
-FROM nginx:mainline
+ARG NGINX_VERSION="1.23.1"
+ARG NGINX_NAME="nginx-${NGINX_VERSION}"
 
-# DEPENDENCIES
+FROM debian as build
+
+ARG NGINX_VERSION
+ARG NGINX_NAME
+ARG CONFIG=" --prefix=/etc/nginx \
+ --sbin-path=/usr/sbin/nginx \
+ --modules-path=/usr/lib/nginx/modules \
+ --conf-path=/etc/nginx/nginx.conf \
+ --http-log-path=/var/log/nginx/node-access.log \
+ --pid-path=/var/run/nginx.pid \
+ --lock-path=/var/run/nginx.lock \
+ --user=nginx --group=nginx \
+ --with-compat \
+ --with-file-aio \
+ --with-threads \
+ --with-http_addition_module \
+ --with-http_auth_request_module \
+ --with-http_dav_module \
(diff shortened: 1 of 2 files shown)


DiegoRBaquero (approved):

gruns:
awesome work @enolal826!! 🙌 🙌 

1. i cant tell from reading https://github.com/google/ngx_brotli whether responses are *always* compressed if the response mimetype matches one in `brotli_types`, or whether responses are only compressed if the response mimetype matches one in `brotli_types` *and* the requesting client announced they support brotli with the `Content-Encoding: br, ...` header

    we only want the latter. the vast majority of browsers support brotli, but ie 11 notably does not https://caniuse.com/brotli

2. add gzip fallback compression when the browser doesn't support brotli. the same mimetype list of compressable filetypes should be used

3. is the default brotli (and gzip) compression level of 6 best for us? probably. but i just want to make sure

gruns:
any way to deduplicate the `gzip_types` and `brotli_types` variables? their values are identical

https://github.com/filecoin-saturn/L1-node/pull/35/commits/b5e1cb2f08275d21a9bac3af766633071280b30f#diff-072cd79f6429749707ffb91c13810f82fa3ef6d0d9a8fddaa7c90e6e5c7bc4aeR43-R52

and

https://github.com/filecoin-saturn/L1-node/pull/35/commits/2f5182e7cbe0c40672db93933ed4ea11dc20465a#diff-072cd79f6429749707ffb91c13810f82fa3ef6d0d9a8fddaa7c90e6e5c7bc4aeR43-R52

Issue #22, opened 2022-05-11
compress compressable filetypes, dont compress everything else


gruns:
* compress these filetypes: https://support.cloudflare.com/hc/en-us/articles/200168396-What-will-Cloudflare-compress-

* dont compress everything else to avoid needless compression on assets already compressed, like jpgs, mp4s, mp3s, etc

* compress either with gzip or brotli (#21), as supported by the client

DiegoRBaquero:
Per top results in google, it makes sense to compress everything rather than not. Bandwidth is more expensive than CPU and we can later optimize by peeking into and compressing only the ones with actual gains.

gruns:
> Per top results in google, it makes sense to compress everything rather than not.

link?

DiegoRBaquero:
https://serverfault.com/questions/253074/what-is-the-best-nginx-compression-gzip-level
https://sudonull.com/post/172479-GZip-and-nginx-performance-impact
http://paulbuchheit.blogspot.com/2009/04/make-your-site-faster-and-cheaper-to.html

PR #39, opened 2022-08-08
Retrieve mime type for CAR files

enoldev:
Second part of #22.

The main idea to get the mime type of the file(s) is to:

1. Iterate over the blocks of the file
2. Index the X first bytes of every block
3. From the root block, traverse the tree to find out the first block of the file(s).

I have considered that the are three types of blocks.

1. `directory`: we're only interested in its `links` (i.e. the files that it points to)
2. `file` with links: in a large file, this blocks acts as a bridge to the data blocks of the file.
In this type of block, we're only interested in its `links` (it might point to another block withs links or actual data blocks)
3. `file` without links: the actual data blocks.

When we traverse the tree, we're looking for the first leftest data block. Then, from this block, we get the magic number of the file.

- The `CARMimeTypeRetriever` class indexes the X first bytes of every block (because we don't know what block is going to be the first leftest)
- The `CARMimeTypeRetriever.addBlock` method parses the bytes of the block by creating a `Block` object. Then, a `BlockIndexEntry` is included in the file (basically, a reduced version of a `Block`, which just contains the X first bytes).
- After iterating over the file, we call the `CARMimeTypeRetriever.retrieveType` method by passing the root CID. This method will iterate over the tree to find the first leftest block in the index.

Once we have the mime type, we need Nginx to conditionally compress the file. If the file has to be compressed, then a `Content-Type: application/vnd.ipld.car ; compress` header will be sent.

Things to consider:
1. Currently we are only considering that a directory can be in the root of the tree. Is it possible to have several directories? For example, a directory pointing to directories?
2. Error management: what happens if we can't determinate the mime type of a file? For example, plain text files won't have a mime type (e.g. txt).

### DIFF
--- a/container/nginx/nginx.conf
+++ b/container/nginx/nginx.conf
@@ -66,7 +66,8 @@ http {
                  application/x-ttf application/otf application/x-otf application/truetype application/opentype
                  application/x-opentype application/font-woff application/eot application/font application/font-sfnt
                  application/wasm application/javascript-binast application/manifest+json application/ld+json
-                 application/graphql+json application/geo+json;
+                 application/graphql+json application/geo+json 'application/vnd.ipld.car ; compress';
+
     # == end custom node directives ==
 
     include /etc/nginx/conf.d/*.conf;
--- a/container/shim/package-lock.json
+++ b/container/shim/package-lock.json
@@ -14,6 +14,8 @@
         "blockstore-core": "^1.0.5",
         "debug": "^4.3.4",
         "express": "^4.18.1",
+        "file-type": "^17.1.6",
+        "ipfs-unixfs": "^6.0.9",
         "ipfs-unixfs-exporter": "^7.0.11",
         "mime-types": "^2.1.35",
         "multiformats": "^9.7.0",
@@ -362,6 +364,11 @@
       "engines": {
         "node": ">=6"
       }
+    },
+    "node_modules/@tokenizer/token": {
+      "version": "0.3.0",
+      "resolved": "https://registry.npmjs.org/@tokenizer/token/-/token-0.3.0.tgz",
+      "integrity": "sha512-OvjF+z51L3ov0OyAU0duzsYuvO01PH7x4t6DJx+guahgTnBHkhJdG7soQeTSFLWN3efnHyibZ4Z8l2EuWwJN3A=="
     },
     "node_modules/@types/json-schema": {
       "version": "7.0.11",
@@ -2485,6 +2492,22 @@
         "node": "^10.12.0 || >=12.0.0"
       }
     },
+    "node_modules/file-type": {
+      "version": "17.1.6",
+      "resolved": "https://registry.npmjs.org/file-type/-/file-type-17.1.6.tgz",
+      "integrity": "sha512-hlDw5Ev+9e883s0pwUsuuYNu4tD7GgpUnOvykjv1Gya0ZIjuKumthDRua90VUn6/nlRKAjcxLUnHNTIUWwWIiw==",
+      "dependencies": {
+        "readable-web-to-node-stream": "^3.0.2",
+        "strtok3": "^7.0.0-alpha.9",
+        "token-types": "^5.0.0-alpha.2"
+      },
+      "engines": {
+        "node": "^12.20.0 || ^14.13.1 || >=16.0.0"
+      },
+      "funding": {
+        "url": "https://github.com/sindresorhus/file-type?sponsor=1"
+      }
+    },
     "node_modules/fill-range": {
       "version": "7.0.1",
       "resolved": "https://registry.npmjs.org/fill-range/-/fill-range-7.0.1.tgz",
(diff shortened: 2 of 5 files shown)


guanzo:
> Is it possible to have several directories? For example, a directory pointing to directories?

Do you mean an array of directories all in one Node? I don't think so, based on the [spec datatypes](https://github.com/ipfs/specs/blob/main/UNIXFS.md#data-format)

Or do you mean the links of the root Node can point to directories? Then yes that's possible.

> Error management: what happens if we can't determinate the mime type of a file? 

Don't set a content-type header and let nginx serve the CAR file as is.

> For example, plain text files won't have a mime type (e.g. txt).

The lib that go-ipfs uses to detect mimetype can apparently detect txt files. https://github.com/gabriel-vasile/mimetype#features
Maybe js has something similar.

Some thoughts:

* [go-ipfs first checks](https://github.com/ipfs/kubo/blob/9bd346e25004df4fd7927ab151097f975fb433c2/core/corehttp/gateway_handler_unixfs_file.go#L54) for the `?filename=<filename>` query and if exists, sets the mime type based on the file extension. Then it falls back to inspecting bytes. Worth implementing in another PR since we're trying to be compatible with the IPFS gateway.

* This appears to read the whole file in memory, detect mime type, and then start streaming to the client. This isn't ideal for streaming. But I also understand that we won't know "what block is going to be the first leftest" unless we have all the blocks. Tricky...

gruns:
> When we traverse the tree, we're looking for the first leftest data block. Then, from this block, we get the magic number of the file.

what about CAR files containing multiple files?

car files we create in the future for saturn will, at least initially, only contain one file. but we, saturn, didn't create all the CAR files that will be requested and thus fetched from the ipfs gateway. thus, car files containing multiple files will flow through saturn as fetched from the ipfs gateway and need to be supported

we discussed a few different compression strategies for car files containing multiple files in DMs. to enumerate a few of those here:

1. dont compress any car files containing multiple files. this is simple but non-optimal but potentially acceptable if car files containing multiple files are a minority of files fetched from the ipfs gateway

2. compress them if *any* files in the car file are compressable. this is non-optimal. eg a 10MB car file with 9.8mb of video data and 200kb of text. it's likely not worth needlessly compressing the entire file to compress 200kb, or 2%, of the file

3. apply some rule of thumb heuristic like if the total size of all compressable content in the car file is more than half the file, then compress it. otherwise, don't. eg:
    * 1mb text, 9mb video -> dont compress
    * 5mb text, 5mb video -> compress
    * 9mb text, 1mb video -> compress
    
>> Error management: what happens if we can't determinate the mime type of a file?
>
> Don't set a content-type header and let nginx serve the CAR file as is.

yep 👍 

> Error management: what happens if we can't determinate the mime type of a file?

is determining the mimetype of a file by its 'magic number', or file signature (https://en.wikipedia.org/wiki/List_of_file_signatures) 

1. robust? ie does it produce a result for 99.9% of files? and
2. reliable? or does it guess the incorrect mimetype some considerable amount of the time?

if mimetype determination by file signature isn't robust enough, we could use a waterfall:

1. determine the mimetype by file signature. eg first X bytes of the file are `89 50 4E 47 0D 0A 1A 0A` -> `image/png`
2. if the mimetype cant be determined by the file signature, fall back to the file extension. eg `foo.png` -> `.png` -> `image/png`
3. if the mimetype cant be determined by file signature nor file extension, don't set the content-type header and let nginx serve the car file as-is

if mimetype determination by file signature isnt reliable enough, we could use the file extension as a second vote of confidence. eg

1. if the file signature mimetype and the file extension mimetype agree: use that mimetype
2. but if the file signature mimetype and the file extension mimetype disagree: dont set the content-header and let nginx serve the car file as-is

note that file signature and file extension are guaranteed not to match some non-trivial percentage of the time. think about how many files on your own filesystem are incorrectly named. like a `foo.png` image which is actually a jpg. happens all the time

gruns:
@enolal826 one thing we could do for this feature is to break it down into smaller, shippable pieces. eg a roadmap to implement iteratively. like:

1. initially only support car files containing a single file and only determine mimetype by file signature
2. then add support for car files containing multiple files
3. then add support for a waterfall of mimetype determination where the file extension is also used (but only do this is mimetype determination by file signature is insufficient)
etc

gruns:
> This appears to read the whole file in memory, detect mime type, and then start streaming to the client. This isn't ideal for streaming. But I also understand that we won't know "what block is going to be the first leftest" unless we have all the blocks. Tricky...

blocks should be freed after being read, and only the first X bytes of data blocks should be cached in memory until the first leftest data block has been read, after which there's no point reading more blocks (in car files containing only one file, at least). where X bytes are all that's needed to determine the mimetype by file signature

at the end of the day, we absolutely cannot read an entire file into memory. otherwise to brick every server one just `curl http://strn.pl/cid/<cid>` a cid larger than the memory of the server and 💥

enolal826:
Ideally, we would start the streaming as soon as we read the blocks, but we find a limitation: it is not possible to set HTTP headers once the streaming has started.

We rely on an HTTP header for NGINX to conditionally compress the file, so if we start the streaming, we won't be able to set the header at the end (after reading the blocks, and determining the type).

PR #40, opened 2022-08-09
Add L1<>L2 interop. Closes #37

### LINKED ISSUE #37: Add GET /register/:L2_ID
L2 nodes will call this route to establish a long-lived connection, so that L1 can on demand make requests to it.

Protocol see https://www.notion.so/pl-strflt/Elaborate-L1-L2-MVP-flow-and-mechanics-793d381d04974a6b903078bf10e72b2a.

For https://github.com/filecoin-saturn/L2-node/issues/48.

juliangruber:
Closes #37.

Documentation of approach and current direction: https://gist.github.com/juliangruber/032276dbdd44e056d5f7fc18d8debc3e

## TODO

- [x] Make the L2 fallback configurable
  - [x] Disabled (main net)
  - [x] Fire and forget (test net initially, then main net)
  - [x] Full integration (post l1 launch)
- [x] Verify L2's car response
- [x] Extract raw data from car if `?format=raw`

Mock client:

```js
import http2 from 'node:http2'
import { CarWriter } from '@ipld/car'
import { Readable } from 'node:stream'
import { CID } from 'multiformats/cid'
import { sha256 } from 'multiformats/hashes/sha2'
import * as raw from 'multiformats/codecs/raw'

const client = http2.connect('https://127.0.0.1')
const req = client.request({
  ':path': `/register/http2?nocache=1`,
})
req.on('data', async d => {
  const line = d.toString().trim()
  if (line === '') return
  const obj = JSON.parse(line)
  console.log(obj)
  if (obj.cid) {
    console.log('-> sending reply')
    const r = client.request({
      ':path': `/data/${obj.cid}`,
      ':method': 'POST'
    })
    const bytes = new TextEncoder().encode('hi')
    const hash = await sha256.digest(raw.encode(bytes))
    const cid = CID.create(1, raw.code, hash)
    const { writer, out } = await CarWriter.create(cid)
    Readable.from(out).pipe(r)
    await writer.put({ cid, bytes })
    await writer.close()
  }
})
req.on('end', () => {
  client.close()
})
req.end()
```

```bash
$ NODE_TLS_REJECT_UNAUTHORIZED=0 NODE_NO_WARNINGS=1 node l2.mjs
```

### DIFF
--- a/container/shim/src/index.js
+++ b/container/shim/src/index.js
@@ -168,6 +168,7 @@ if (cluster.isPrimary) {
     if (carResponse) {
       try {
         await pipeline(carResponse.req, res)
+        return
       } finally {
         carResponse.res.end()
       }
--- a/container/shim/src/index.js
+++ b/container/shim/src/index.js
@@ -224,7 +224,7 @@ async function maybeRespondFromL2 (req, res, { cid, format }) {
   return false
 }
 
-app.get('/register/:l2NodeId', async function (req, res) {
+app.get('/register/:l2NodeId', asyncHandler(async function (req, res) {
   res.writeHead(200, {
     'Cache-Control': 'no-cache'
   })
@@ -245,7 +245,7 @@ app.get('/register/:l2NodeId', async function (req, res) {
   if (!cleanedUp.value) {
     removeConnectedL2Node(l2NodeId)
   }
-})
+}))
 
 app.post('/data/:cid', function (req, res) {
   const { cid } = req.params
--- a/container/shim/src/index.js
+++ b/container/shim/src/index.js
@@ -10,26 +10,20 @@ import xorDistance from 'xor-distance'
 import pDefer from 'p-defer'
 import pTimeout from 'p-timeout'
 import timers from 'node:timers/promises'
-import parseArgs from 'minimist'
+import asyncHandler from 'express-async-handler'
 
-import { addRegisterCheckRoute, deregister, register } from './modules/registration.js'
+import { addRegisterCheckRoute } from './modules/registration.js'
 import {
-  FIL_WALLET_ADDRESS,
   IPFS_GATEWAY_ORIGIN,
   L2_FIRE_AND_FORGET,
-  NODE_OPERATOR_EMAIL,
   NODE_UA,
   NODE_VERSION,
   nodeId,
-  PORT,
   SATURN_NETWORK,
   TESTING_CID
 } from './config.js'
 import { streamCAR, streamRawFromCAR } from './utils/car.js'
-import { trapServer } from './utils/trap.js'
 import { debug } from './utils/logging.js'
-
-import { submitRetrievals, initLogIngestor } from './modules/log_ingestor.js'
 
 const { https } = followRedirects
 
@@ -47,31 +41,6 @@ const PROXY_RESPONSE_HEADERS = [
   'x-ipfs-datasize',
   'x-content-type-options'
 ]
-
-const argv = parseArgs(process.argv.slice(2))
-
-debug('Saturn L1 Node')
-debug.extend('id')(nodeId)
-debug.extend('version')(NODE_VERSION)
-debug.extend('important')('===== IMPORTANT =====')
-debug.extend('important')(`Earnings will be sent to Filecoin wallet address: ${FIL_WALLET_ADDRESS}`)
-debug.extend('important')(NODE_OPERATOR_EMAIL ? `Payment notifications and important update will be sent to: ${NODE_OPERATOR_EMAIL}` : 'NO OPERATOR EMAIL SET, WE HIGHLY RECOMMEND SETTING ONE')
-debug.extend('important')('===== IMPORTANT =====')
-
-process.on('SIGQUIT', shutdown)
-process.on('SIGINT', shutdown)
-
-setTimeout(async function () {
-  if (argv.register !== false) {
-    await register(true).catch(err => {
-      debug(`Failed to register ${err.name} ${err.message}`)
-      process.exit(1)
-    })
-  }
-
-  // Start log ingestor
-  await initLogIngestor()
-}, 500)
 
 const ipfsAgent = new https.Agent({
   keepAlive: true,
@@ -104,11 +73,7 @@ app.get('/favicon.ico', (req, res) => {
   res.sendStatus(404)
 })
 
-// Whenever nginx doesn't have a CAR file in cache, this is called
-app.get('/ipfs/:cid', handleCID)
-app.get('/ipfs/:cid/:path*', handleCID)
-
-async function handleCID (req, res) {
+const handleCID = asyncHandler(async (req, res) => {
   const cid = req.params.cid
   const format = getResponseFormat(req)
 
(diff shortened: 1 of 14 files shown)


container/shim/src/index.js
```diff
@@ -3,6 +3,12 @@ import { cpus } from 'node:os'
 import express from 'express'
 import mimeTypes from 'mime-types'
 import followRedirects from 'follow-redirects'
+import crypto from 'node:crypto'
```
bajtos, line 6: Yay for using `node:` prefix for built-in modules 👍🏻

container/shim/src/index.js
```diff
@@ -42,161 +47,231 @@ const PROXY_RESPONSE_HEADERS = [
...
+})
 
-  setTimeout(async function () {
-    if (argv.register !== false) {
-      await register(true).catch(err => {
-        debug(`Failed to register ${err.name} ${err.message}`)
-        process.exit(1)
-      })
-    }
+const app = express()
```
bajtos, line 78: Express is known to have a pretty bad performance, plus it's not very actively maintained. I strongly recommend replacing it with a more modern alternative like  [Fastify](https://www.fastify.io).

Having said that, such change is most likely out of scope of this pull request.

container/shim/src/index.js
```diff
@@ -42,161 +47,231 @@ const PROXY_RESPONSE_HEADERS = [
...
 
-    // Testing CID
-    if (cid === TESTING_CID) {
-      return res.send(testCAR)
+  let carResponse
+  try {
+    carResponse = await pTimeout(onResponse.promise, {
+      milliseconds: 10_000
+    })
+  } catch {}
```
bajtos, line 163: Should we print a debug log when an error is caught?

container/shim/src/index.js
```diff
@@ -42,161 +47,231 @@ const PROXY_RESPONSE_HEADERS = [
...
+  if (carResponse) {
+    try {
+      await pipeline(carResponse.req, res)
```
guanzo, line 166: We should incrementally verify the CAR file as we stream to the client, like we do with the gateway response. https://github.com/filecoin-saturn/L1-node/blob/9834d819703619d012f72370deb9bdd8c647ebf2/container/shim/src/index.js#L205

juliangruber: Great point, I'll add that

container/shim/src/index.js
```diff
@@ -42,161 +47,231 @@ const PROXY_RESPONSE_HEADERS = [
...
+app.get('/ipns/:cid/:path*', handleCID)
+app.get('/ipfs/:cid', handleCID)
+app.get('/ipfs/:cid/:path*', handleCID)
 
-  // Whenever nginx doesn't have a CAR file in cache, this is called
-  app.get('/ipns/:cid', handleCID)
-  app.get('/ipns/:cid/:path*', handleCID)
-  app.get('/ipfs/:cid', handleCID)
-  app.get('/ipfs/:cid/:path*', handleCID)
+async function handleCID (req, res) {
```
bajtos, line 105: Express does not support async request handlers (yet another reason why to use Fastify).

If this async function returns a promise that's eventually rejected, unhandled promise rejection is triggered at the process level. I don't remember whether this prints a warning or crashes the process. I think neither is desirable.

A possible solution: convert the async hander into callback-based.

```js
function wrap(fn) {
  return function(req, res, next) {
    fn(req, res)
      .catch(next)
      .catch(
        // crash if callback throws
        err => process.nextTick(() => { throw err; })
      });
  };
}

const handleCID = wrap(async function handleCIDimpl(req, res) {
  // your code
}
```

container/shim/src/index.js
```diff
@@ -42,161 +47,231 @@ const PROXY_RESPONSE_HEADERS = [
...
+}
 
-    req.on('close', () => {
-      clearTimeout(timeout)
-      if (!res.writableEnded) {
-        debug.extend('error')('Client aborted early, terminating gateway request')
-        ipfsReq.destroy()
-      }
-    })
+app.get('/register/:l2NodeId', async function (req, res) {
```
bajtos, line 232: Ditto, express does not support async functions for request handlers.

bajtos (commented):
Nice! I left a few comments to consider.

container/shim/src/index.js
guanzo: this package is often recommended to handle this issue https://www.npmjs.com/package/express-async-errors

bajtos: Quoting from https://www.fastify.io/benchmarks/
- Fastify - 53913 req/sec
- Express - 11739 req/sec

On the second thought, I think the Node.js HTTP framework is not a bottleneck in our current L1 implementation since most CPU is consumed by Nginx anyways.

I am fine to stay with Express for now. To me, it's important to have robust & reliable error handling. This can be achieved with Express, too: using one of the plugins like the one mentioned above, wrapping our handler to convert them to callback style, or we can even rewrite the handlers into callback style.

container/shim/src/index.js
juliangruber: I agree, we have an open issue for this: https://github.com/filecoin-saturn/L1-node/issues/33

container/shim/src/index.js
juliangruber: Not sure, it would print one for every cache miss 🤔 We already have a line for this below: `debug(`Fetch ${req.path} from IPFS`)`, and the only error that can be thrown here is a timeout.

container/shim/src/index.js
juliangruber: I agree, and captured this in https://github.com/filecoin-saturn/L1-node/issues/33, which we should also resolve before launch.

container/shim/src/index.js
juliangruber: https://github.com/filecoin-saturn/L1-node/pull/48

container/shim/src/index.js
```diff
@@ -42,159 +48,246 @@ const PROXY_RESPONSE_HEADERS = [
...
+      res.write(`${JSON.stringify(payload)}\n`)
+    })
+  if (L2_FIRE_AND_FORGET) {
+    return false
+  }
 
-      res.status(statusCode)
-      proxyResponseHeaders(fetchRes, res)
+  const onResponse = pDefer()
+  openCARRequests.set(cid, onResponse)
```
guanzo, line 225: The logic only handles cids, I think we also need to handle file paths in the L2 request payload, cache keys (`openCARRequests`), endpoints (`app.post('/data/:cid'`), etc.

Currently if I request https://strn.pl/ipfs/QmWKDHujv2nNZkBaxj7UvuraoVzgH79pbZqxGyy89e4QnM/999.json, I'd receive https://strn.pl/ipfs/QmWKDHujv2nNZkBaxj7UvuraoVzgH79pbZqxGyy89e4QnM instead. The former is a file, the latter is a directory.

guanzo (changes requested):

container/shim/src/index.js
juliangruber: Good catch!

juliangruber: Asked for clarification in https://filecoinproject.slack.com/archives/C031K3BG527/p1661250640769579

juliangruber: as discussed, I've disabled l2 fallback

guanzo (approved):

DiegoRBaquero (approved):
Should be fine to deploy given that it´s FF for test net.

container/shim/src/index.js
bajtos: Fair enough 👍🏻

bajtos (approved):

PR #41, opened 2022-08-10
proxy /ipns to the ipfs gateway, bypassing the shim altogether

guanzo:
/ipns requests bypass the shim. Requests are still logged and reported.

### DIFF
--- a/container/nginx/confs/proxy.tlsconf
+++ b/container/nginx/confs/proxy.tlsconf
@@ -50,6 +50,21 @@ server {
         return 302 https://strn.network;
     }
 
+    location /ipns/ {
+        proxy_pass  https://ipfs.io;
+
+        if ($request_method = 'OPTIONS') {
+            add_header 'Timing-Allow-Origin'            '*';
+            add_header 'Access-Control-Allow-Origin'    '*';
+            add_header 'Access-Control-Allow-Methods'   'GET, POST, OPTIONS';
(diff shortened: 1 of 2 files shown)


container/nginx/confs/proxy.tlsconf
```diff
@@ -50,6 +50,21 @@ server {
         return 302 https://strn.network;
     }
 
+    location /ipns/ {
```
DiegoRBaquero, line 53: Shouldn't we rewrite redirections? so once ipfs gateway redirs ipns to ipfs CID, it's a saturn one

guanzo: /ipns endpoints don't return redirections, the gateway resolves the cid internally and returns the content. 

`curl 'https://ipfs.io/ipns/tokens.uniswap.org' -v -o /dev/null -k`

container/nginx/confs/proxy.tlsconf
```diff
@@ -50,6 +50,21 @@ server {
         return 302 https://strn.network;
     }
 
+    location ~ ^/(ipns|api)/ {
```
DiegoRBaquero, line 53: I'd like to see even a super low cache 1m

guanzo: I think thats reasonable.  will make it a TODO

PR #52, opened 2022-08-25
Remove express. Closes #33

### LINKED ISSUE #33: Remove express

juliangruber:
Closes #33.

Motivation:
- express is outdated
- L1 doesn’t need high level web features
- This should be the fastest route perf wise, since it’s mostly using node core. At the same time, the complexity is kept low (in my mind)

Prior art, in heavy production use: https://github.com/electron/update.electronjs.org/blob/master/index.js

## TODO

- [ ] write tests

### DIFF
--- a/container/shim/package-lock.json
+++ b/container/shim/package-lock.json
@@ -13,13 +13,13 @@
         "@multiformats/blake2": "^1.0.10",
         "blockstore-core": "^1.0.5",
         "debug": "^4.3.4",
-        "express": "^4.18.1",
         "follow-redirects": "^1.15.1",
+        "http-assert": "^1.5.0",
         "ipfs-unixfs-exporter": "^7.0.11",
         "mime-types": "^2.1.35",
         "minimist": "^1.2.6",
         "multiformats": "^9.7.0",
-        "node-fetch": "^3.2.6",
+        "node-fetch": "^3.2.10",
         "pretty-bytes": "^6.0.0"
       },
       "devDependencies": {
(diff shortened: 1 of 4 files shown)


juliangruber:
I'm looking for early stage review on the general approach and implementation. Before merging this we need to have a basic test suite, which I'm planning to add next, independently.

juliangruber:
I'm looking for early stage review on the general approach and implementation. Before merging this we need to have a basic test suite, which I'm planning to add next, independently.

guanzo:
Looks fine to me, easy to follow and basing it off reputable prior art was a good move.

I'd like to see https://github.com/filecoin-saturn/L1-node/pull/40 merged first, the code has diverged a bit (this PR still uses cluster).

bajtos (approved):
Nice! I like the idea of removing Express from our stack.

In general, I would be reluctant to build directly on top of the Node.js core `http` module because it misses many features required for any sensible API or Web server - CORS, request body parsing, and so on.

However, our situation is different - most of the complicated stuff is handled by Nginx so that L2 can be bare-bone. Avoiding any HTTP framework is a good idea 👍🏻 

I feel the file `container/shim/src/index.js` is getting too large; it implements too many different things. It would be great to split it into multiple smaller files. For example, we can start with extracting `class L1Node`. This is nice to have and can be done in a different pull request (or even not at all).

juliangruber:
> In general, I would be reluctant to build directly on top of the Node.js core `http` module because it misses many features required for any sensible API or Web server - CORS, request body parsing, and so on.
> 
> However, our situation is different - most of the complicated stuff is handled by Nginx so that L2 can be bare-bone. Avoiding any HTTP framework is a good idea 👍🏻

I wouldn't inline much HTTP functionality anyway, whatever we can we should source from npm. cors and body parsing for example are easily available as independent modules.

> I feel the file `container/shim/src/index.js` is getting too large; it implements too many different things. It would be great to split it into multiple smaller files. For example, we can start with extracting `class L1Node`. This is nice to have and can be done in a different pull request (or even not at all).

I very much agree, and believe we're already there once I merge `main` into this branch, since we already split the request handling logic from the http server / cluster / registration stuff.

PR #54, opened 2022-08-29
Separate http library from server

juliangruber:
In preparation for adding tests in https://github.com/filecoin-saturn/L1-node/issues/49, separate the request handler from the server logic, so that the business logic can be tested in isolation.


juliangruber:
This is a bit difficult to review, because of the indentation change, but hopefully through this patch we'll have less indentation changes in the future

DiegoRBaquero (approved):

PR #55, opened 2022-08-31
Add basic test suite. Closes #49

### LINKED ISSUE #49: Create test suite

juliangruber:
This PR adds a basic test suite, for the most important paths. It tests the HTTP server on the HTTP level, and mocks external resources.

I left a bunch of `todos` for tests that can be added in the future, and there are even more, but I first wanted to land the basic infrastructure and have a stable iteration platform.

The test runner used is [`test`](https://github.com/nodejs/node-core-test), which is a user-land port of the currently experimental [node core test runner](https://nodejs.org/api/test.html). I like using this module since it helps improve the node core module, and I think it will be very useful to have a solid test runner in node core.

Closes #49.

### DIFF
--- a/container/shim/package.json
+++ b/container/shim/package.json
@@ -3,7 +3,7 @@
   "type": "module",
   "scripts": {
     "prepare": "cd ../.. && husky install container/shim/.husky",
-    "test": "eslint --fix ."
+    "test": "eslint --fix . && FIL_WALLET_ADDRESS=test NODE_OPERATOR_EMAIL=test node--test"
   },
   "dependencies": {
     "@ipld/car": "^4.1.3",
@@ -35,7 +35,8 @@
     "eslint-plugin-jsdoc": "^39.3.3",
     "eslint-plugin-promise": "^6.0.0",
     "husky": "^8.0.1",
-    "nodemon": "^2.0.18"
+    "nodemon": "^2.0.18",
+    "test": "^3.2.1"
   },
   "eslintConfig": {
     "extends": "ipfs",
--- a/container/shim/package.json
+++ b/container/shim/package.json
@@ -37,6 +37,7 @@
     "eslint-plugin-jsdoc": "^39.3.3",
     "eslint-plugin-promise": "^6.0.0",
     "husky": "^8.0.1",
+    "nock": "^13.2.9",
     "nodemon": "^2.0.18",
     "test": "^3.2.1"
   },
--- a/container/shim/package.json
+++ b/container/shim/package.json
@@ -5,7 +5,7 @@
     "prepare": "cd ../.. && husky install container/shim/.husky",
     "test": "npm run test:lint && npm run test:unit",
     "test:lint": "eslint --fix .",
-    "test:unit": "FIL_WALLET_ADDRESS=test NODE_OPERATOR_EMAIL=test node--test"
+    "test:unit": "FIL_WALLET_ADDRESS=test NODE_OPERATOR_EMAIL=test L2_FIRE_AND_FORGET=true node--test"
   },
   "dependencies": {
     "@ipld/car": "^4.1.3",
--- a/container/shim/package.json
+++ b/container/shim/package.json
@@ -3,7 +3,9 @@
   "type": "module",
   "scripts": {
     "prepare": "cd ../.. && husky install container/shim/.husky",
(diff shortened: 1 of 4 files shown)


container/shim/src/config.js
```diff
@@ -15,7 +15,9 @@ export const NODE_OPERATOR_EMAIL = process.env.NODE_OPERATOR_EMAIL || error('NOD
 export const IPFS_GATEWAY_ORIGIN = process.env.IPFS_GATEWAY_ORIGIN || 'https://ipfs.io'
 export const TESTING_CID = 'QmXjYBY478Cno4jzdCcPy4NcJYFrwHZ51xaCP8vUwN9MGm'
 export const nodeId = await readFile(NODE_ID_FILE_PATH, 'utf-8').catch(() => false) || createNodeId()
-export const L2_FIRE_AND_FORGET = SATURN_NETWORK === 'test'
+export const L2_FIRE_AND_FORGET = process.env.L2_FIRE_AND_FORGET
+  ? process.env.L2_FIRE_AND_FORGET === 'true'
+  : SATURN_NETWORK === 'test'
```
juliangruber, line 20: this is a it awkward, and unflexible. What do you think about adding a `config` option for creating that express app, which lets us override values without needing to change `process.env`?

container/shim/test/test.js
```diff
@@ -0,0 +1,101 @@
...
+        Buffer.from(await (await res.blob()).arrayBuffer()),
+        testCAR
+      )
+    })
+    await t.test('response headers', async t => {
+      const res = await fetch(`${address}/ipfs/${TESTING_CID}`)
+      assert.strictEqual(res.headers.get('content-type'), 'application/octet-stream')
+      assert.strictEqual(res.headers.get('cache-control'), 'public, max-age=31536000, immutable')
+      assert.strictEqual(res.headers.get('saturn-node-id'), nodeId)
+      assert.strictEqual(res.headers.get('saturn-node-version'), DEV_VERSION)
```
juliangruber, line 60: I don't want to just repeat these static headers, so that the test suite is perceived as an unnecessary crutch impeding progress, however I want to test the L1 node's public interface, so we notice when we are breaking its promise.

container/shim/src/config.js
DiegoRBaquero: I am not exactly sure what you mean by this, do you have an example?

DiegoRBaquero (approved):

container/shim/package.json
```diff
@@ -35,7 +37,9 @@
     "eslint-plugin-jsdoc": "^39.3.3",
     "eslint-plugin-promise": "^6.0.0",
     "husky": "^8.0.1",
-    "nodemon": "^2.0.18"
+    "nock": "^13.2.9",
+    "nodemon": "^2.0.18",
+    "test": "^3.2.1"
```
DiegoRBaquero, line 42: Any reason not to use native test? https://nodejs.org/api/test.html

DiegoRBaquero: Nvm, still experimental, and Node 18+ only

container/shim/src/config.js
juliangruber:
```js
// in tests:

const app = createApp({
  l2FireAndForget: true,
  filWalletAddress: 'test'
})

// in production

const app = createApp({
  l2FireAndForget: SATURN_NETWORK === 'test',
  filWalletAddress: process.env.FIL_WALLET_ADDRESS || error('oh no')
})
```

juliangruber: so basically we let `config.js` be passed as an object, and when the server is launched through `bin/shim.js` we populate config from env vars

container/shim/package.json
juliangruber: that's it 👍

container/shim/src/config.js
DiegoRBaquero: We could split `config.js` into `env.js` and `config.js` where `env.js` only has extraction from `process.env` and it's config that has an exported class that allows constructor overrides, wdyt?

bajtos (approved):
LGTM as far as I am concerned.

container/shim/src/config.js
juliangruber: I gave this a shot, but quickly ran into limitations of the current setup. I'm going to revisit this after https://github.com/filecoin-saturn/L1-node/pull/52, when L1 was turned into a class.

juliangruber:
Hindsight please @guanzo if you want 🙏

PR #57, opened 2022-09-05
update nginx max body size to 10g

juliangruber:
@DiegoRBaquero do we need to limit this to the `POST /data/` route as well? I wasn't able to figure this part out (my nginx cookbook is at home :|)


DiegoRBaquero (approved):
I think here is fine for now as there isn't a more specific block.

PR #62, opened 2022-09-10
Run Kubo Gateway tests on the L1

guanzo:
[Kubo](https://github.com/ipfs/kubo) has a series of tests that test the IPFS HTTP Gateway functionality. By successfully running these tests on the L1, it proves that the L1 can correctly handle requests for https://ipfs.io. 

The goal is to be API compatible with the [Path Gateway spec](https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md). Some of the tests are concerned with the other gateway modes (such as subdomains and dnslink), which the L1 doesn't support. In order to keep the tests focused on Path Gateway correctness, Saturn will maintain a [kubo fork](https://github.com/filecoin-saturn/kubo/tree/saturn) with minimal test modifications. Forking was also necessary to insert the L1 node into the tests.

Kubo test: `test client` -> `ipfs daemon`
Fork test:   `test client` -> `L1 node` -> `ipfs daemon`

Notable changes:
* Allow non-tls L1 to run with all endpoints enabled. Previously, if tls certs didn't exist then the `/ipfs` endpoint would be inactive. If tls certs exist, then the http -> https redirection will still occur. 
  *  This was implemented so the tests could continue to send http requests. It also makes L1 dev a little more convenient because you aren't required to have certs anymore.
  * There should be no negative effects in production. The Orchestrator still checks for a valid cert before letting an L1 into the network.
* Return 301 redirects to the client instead of following them. This should be a low impact change as redirects are a small percentage of requests
* 2 new environment variables 
  * `IPFS_GATEWAY_ORIGIN`
    * Default: `https://ipfs.io`
    * Example: `IPFS_GATEWAY_ORIGIN=http://localhost:3245`
  * `ORCHESTRATOR_REGISTRATION`
    * Disables orchestrator registration so the L1 can run in isolation.
    * This replaces the `--no-register` cli flag, as it was a pain to pass to the node instance.
    * Default: `true`
    * Example: `ORCHESTRATOR_REGISTRATION="false"`
* L1 deployment is contingent on successful Kubo tests.


.github/workflows/release.yml
```diff
@@ -13,7 +14,58 @@ env:
...
 jobs:
+  test-kubo:
+    name: 'test:kubo'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          path: L1-node
+
+      - name: Checkout Kubo fork
```
juliangruber, line 25: I think this needs a docs/ page, describing how the maintenance of the Kubo fork is done. For example, what are the tasks for keeping it up to date, and how often does one do this?

juliangruber, line 25:
```suggestion
      - name: Check out Kubo fork
```

nit: afaik "checkout" isn't a verb

juliangruber: It would also be useful to document how to run these Kubo tests locally, for faster iterations

.github/workflows/release.yml
```diff
@@ -13,7 +14,58 @@ env:
...
+              echo Skip $file
+              continue
+            fi
+            echo Running $file
+            bash $file
+          done
+
+      # - name: Print L1 logs
+      #   if: always()
+      #   run: cat /tmp/node.log
```
juliangruber, line 65: instead of commenting out these, we could also always attach the logs to the workflow run (I don't think there's harm in doing it every time), using 

```yaml
name: Attach produced packages to Github Action
        uses: actions/upload-artifact@v2
        with:
          name: kubo-logs
          path: /tmp/node.log
```

.github/workflows/release.yml
```diff
@@ -13,7 +14,58 @@ env:
...
+      - name: Run Kubo gateway tests
+        run: |
+          export L1_NODE_REPO_PATH=$GITHUB_WORKSPACE/L1-node
+          #export L1_NODE_LOG_FILE=/tmp/node.log
+
+          cd $GITHUB_WORKSPACE/kubo/test/sharness
+
+          for file in $(find .  -maxdepth 1 -name 't011*-gateway*.sh' -type f); do
+            echo
+            if [[ $file == *"t0111"* || $file == *"t0114"* ]]; then
```
juliangruber, line 55: Why are we skipping these tests?

.github/workflows/release.yml
```diff
@@ -13,7 +14,58 @@ env:
...
+              echo Skip $file
+              continue
+            fi
+            echo Running $file
+            bash $file
+          done
```
juliangruber, line 61: Do you know if shell scripts here have `-x` or something like that set, so they fail properly?

container/shim/src/index.js
```diff
@@ -119,27 +134,36 @@ const handleCID = asyncHandler(async (req, res) => {
...
-  const ipfsUrl = new URL(IPFS_GATEWAY_ORIGIN + req.path)
+  const ipfsUrl = new URL(IPFS_GATEWAY_ORIGIN + toUtf8(req.path))
   if (format) {
     ipfsUrl.searchParams.set('format', format)
   }
-  for (const key of ['filename', 'download']) {
-    if (key in req.query) {
-      ipfsUrl.searchParams.set(key, req.query[key])
-    }
+  for (const [key, val] of Object.entries(req.query)) {
```
juliangruber, line 151: should be safe to pass on I think?

container/shim/src/index.js
```diff
@@ -268,7 +288,21 @@ app.post('/data/:cid', function (req, res) {
...
 addRegisterCheckRoute(app)
 
-export default app
+function proxyRequestHeaders (reqHeaders) {
+  const headers = { 'User-Agent': NODE_UA }
+
+  for (const key of PROXY_REQUEST_HEADERS) {
+    if (key in reqHeaders) {
+      if (key === 'x-if-none-match') {
+        headers['if-none-match'] = reqHeaders[key]
```
juliangruber, line 297: Would there be value in writing this generic for all `x-*` headers?

juliangruber (changes requested):
Nice work, Eric!

.github/workflows/release.yml
guanzo: Skipped t0111-gateway-writeable.sh because L1 doesn't support writeable gateway.
Skipped t0114-gateway-subdomains.sh because L1 doesn't support subdomains.

I'll document this in the workflow

.github/workflows/release.yml
guanzo: From my testing, the workflow step will fail if any of the tests fail. 
I can make it more explicit with `bash $file || exit 1;`

.github/workflows/release.yml
guanzo: True. I could also just uncomment it? We don't really need these logs in long term storage. If someone needs to view the logs it'll be because the workflow failed and they want to view the logs immediately instead of later.

container/shim/src/index.js
guanzo: what do you mean?

container/shim/src/index.js
guanzo: Eh I thought about it, but then we'd proxy headers that shouldn't be proxied, like these two https://github.com/filecoin-saturn/L1-node/blob/c74ebd4359d065b08d047c2dd3df7e4b0a57f386/container/nginx/confs/proxy.tlsconf#L84-L85

This is a hacky workaround since nginx swallows the `if-none-match` header, if we need to forward additional headers in this manner then we can come up with something generic.

.github/workflows/release.yml
guanzo: agreed. working on this atm

.github/workflows/release.yml
juliangruber: Sure!

.github/workflows/release.yml
juliangruber: If that's what it does then I think we can rely on the convention without cluttering things up

container/shim/src/index.js
juliangruber: Before this change, only some query parameters were passed on, now all are passed on. I was wondering if it is safe to do this 🤔

guanzo: hmm yeah, I think it's fine. What's the worse that could happen? (famous last words)

.github/workflows/release.yml
guanzo: Nvm turns out `bash $file || exit 1;` is needed.

container/shim/src/index.js
juliangruber: Agreed! :D

.github/workflows/release.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches: ['main']
     tags: ['*']
     paths:
+      - '.github/**'
```
DiegoRBaquero, line 8: Why would we want to create a new docker image version whenever `.github` files are changed?

container/nginx/confs/non_tls_proxy.conf
```diff
@@ -0,0 +1,9 @@
+server {
+    listen 80 default_server reuseport;
+    listen 8080 default_server reuseport;
```
DiegoRBaquero, line 3:
```suggestion
```

We can remove 8080 now that 80-only is deployed everywhere

container/shim/src/index.js
```diff
@@ -89,7 +104,7 @@ const handleCID = asyncHandler(async (req, res) => {
 
   res.set({
     'Content-Type': mimeTypes.lookup(req.path) || 'application/octet-stream',
-    'Cache-Control': 'public, max-age=31536000, immutable',
+    'Cache-Control': 'public, max-age=29030400, immutable',
```
DiegoRBaquero, line 107: Why lower it?

.github/workflows/release.yml
guanzo: Oops, that's a leftover line from my testing. I'll remove.

container/shim/src/index.js
guanzo: Just so the tests pass. There's not a significant difference between 336 days and 365 days anyhow.

https://github.com/ipfs/kubo/blob/master/test/sharness/t0116-gateway-cache.sh#L74
https://github.com/ipfs/kubo/blob/master/test/sharness/t0117-gateway-block.sh#L71
https://github.com/ipfs/kubo/blob/master/test/sharness/t0118-gateway-car.sh#L132

DiegoRBaquero: I wonder why they chose 336 days, but yeah, it is fine

guanzo:
Can we get this merged.

DiegoRBaquero:
Go for it, squash merging

DiegoRBaquero (approved):

PR #64, opened 2022-09-14
Add HTTP3 support

joaosa:
- Changes
   - Logged http3 usage to the logs
   - Configured nginx to use, upgrade (through `alt-svc`) and report http3 usage (`quic-status` header)
   - Added http3 support to the Dockerfile
      - Built nginx from source (plus boringssl), so we can get tip fixes ASAP
         - I should note these changes were heavily based on https://github.com/macbre/docker-nginx-http3, so kudos to them!
      - Copied the binary over the final image
      - Enabled `--with-http_v3_module` and `--with-stream_quic_module`
   - Validated with `curl -I --http3 https://localhost` and `curl -I --alt-svc alt-svc.txt https://localhost
` and I got http3 back
      - generated my own certs with [mkcert](https://github.com/FiloSottile/mkcert)
      - modified `proxy.tlsconf` to return 200 on `/` instead of
        redirecting to https://strn.network
      - ran nginx without the shim and copied the conf directly `RUN cp /etc/nginx/confs/* /etc/nginx/conf.d/`
      - got the results I wanted
      - reverted these changes

Refer https://quic.nginx.org/readme.html for more info on the config changes.


container/nginx/confs/proxy.tlsconf
```diff
@@ -20,6 +20,7 @@ map $format_query $format_final {
 
 server {
     listen 443 default_server ssl http2 reuseport;
+    listen 443 http3 reuseport;
```
AmeanAsad, line 23: I might be wrong, but isn't `reuseport` only used once per address/port combination (ie we can use it on both the listen directives for port 443, only one) ? 

[Source for this:](http://nginx.org/en/docs/http/ngx_http_core_module.html#listen) 

> The listen directive can have several additional parameters specific to socket-related system calls. These parameters can be specified in any listen directive, but only once for a given address:port pair.

container/nginx/confs/proxy.tlsconf
```diff
@@ -37,6 +38,12 @@ server {
 
     resolver    8.8.8.8 1.1.1.1 valid=600s;
 
+    # Add Alt-Svc header to negotiate HTTP/3.
+    add_header Alt-Svc 'h3=":443"; ma=86400';
```
AmeanAsad, line 42: One thing to note is that currently http3 has multiple draft versions. Clients currently opt in to which draft range they support. Their is currently a draft range that is highly supported and I believe we have to specify those draft numbers in our `Alt-Svc` header in order to use HTTP3 with a client that only supports a specific draft range. 

[Source 1 (Draft compatibility):](https://hacks.mozilla.org/2021/04/quic-and-http-3-support-now-in-firefox-nightly-and-beta/)

> Both the client and server must support the same QUIC and HTTP/3 draft version to connect with each other. For example, Firefox currently supports drafts 27 to 32 of the specification, so the server must report support of one of these versions (e.g., “h3-32”) in Alt-Svc or HTTPS record for Firefox to try to use QUIC and HTTP/3 with that server. Otherwise, we'd be missing out on a bunch of clients that support http3 but we won't be able to serve anything because they only support a specific draft range that we didn't specify in our `Alt-svc` header.

[Source 2 (Nginx Draft range support): ](https://www.nginx.com/blog/our-roadmap-quic-http-3-support-nginx/)
> We implemented multiple draft versions of HTTP/3, from [draft 23](https://datatracker.ietf.org/doc/draft-ietf-quic-http/23/) through [draft 29](https://datatracker.ietf.org/doc/draft-ietf-quic-http/29/) (the most widely implemented version today), and of course the published standard ([draft 34](https://datatracker.ietf.org/doc/draft-ietf-quic-http/34/)).

I'll note that both my sources were published in 2021, so maybe the draft support is much less now and more people are just using the latest standard. 

[Here] is an example of how I added draft range support in the Nginx-Quic image I created while testing.
`add_header alt-svc 'h3-27=":443"; ma=86400, h3-28=":443"; ma=86400, h3-29=":443"; ma=86400, h3=":443"; ma=86400';`

container/nginx/confs/proxy.tlsconf
DiegoRBaquero: why not put `http3` in line 22? I'm not sure if two listen directives can have the same port for the same IPv4 interface.

container/nginx/nginx.conf
```diff
@@ -17,7 +17,7 @@ http {
 
     log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                       '$status $body_bytes_sent "$http_referer" '
-                      '"$http_user_agent" "$http_x_forwarded_for"';
+                      '"$http_user_agent" "$http_x_forwarded_for" "$http3"';
```
DiegoRBaquero, line 20: Keep in mind this log format is not used.

container/nginx/confs/proxy.tlsconf
joaosa: good point @AmeanAsad!

@DiegoRBaquero I initially did that, but I got `nginx: [emerg] "listen" directive "ssl" parameter is incompatible with "http3" in /etc/nginx/conf.d/proxy.tlsconf:22`.
Taking https://quic.nginx.org/readme.html, as an example, I'm thinking this section should be:

```
listen 443 default_server ssl http2;
listen 443 http3 reuseport;
```

container/nginx/confs/proxy.tlsconf
joaosa: That is a sensible point @AmeanAsad. I was initially using it, but ended up cleaning it up after checking https://quic.nginx.org/readme.html. I'll add your header for compatibility with older clients 👍

container/nginx/nginx.conf
joaosa: I did miss that 😅 . thanks!

container/nginx/confs/proxy.tlsconf
DiegoRBaquero: If the parameters are shared for all listen directives with same `address:port` then we should set `reuseport` in the first one (I don't know if it matters, just rather waterfall than only activated for http3, as CPU would probably spike BAD without it in http2)

thinking of:
```
listen 443 default_server ssl http2 reuseport;
listen 443 http3;
```

container/nginx/nginx.conf
```diff
@@ -33,7 +33,7 @@ http {
 
     reset_timedout_connection   on;
 
-    log_format node 'ua=$http_user_agent&&r=$uri&&rid=$request_id&&args=$args&&s=$status&&b=$bytes_sent&&range=$http_range&&lt=$time_iso8601&&rt=$request_time&&addr=$remote_addr&&ff=$http_x_forwarded_for&&ref=$http_referer&&uct=$upstream_connect_time&&uht=$upstream_header_time&&urt=$upstream_response_time&&ucs=$upstream_cache_status';
+    log_format node 'ua=$http_user_agent&&r=$uri&&rid=$request_id&&args=$args&&s=$status&&b=$bytes_sent&&range=$http_range&&lt=$time_iso8601&&rt=$request_time&&addr=$remote_addr&&ff=$http_x_forwarded_for&&ref=$http_referer&&uct=$upstream_connect_time&&uht=$upstream_header_time&&urt=$upstream_response_time&&ucs=$upstream_cache_status $http3';
```
DiegoRBaquero, line 36:
```suggestion
    log_format node 'ua=$http_user_agent&&r=$uri&&rid=$request_id&&args=$args&&s=$status&&b=$bytes_sent&&range=$http_range&&lt=$time_iso8601&&rt=$request_time&&addr=$remote_addr&&ff=$http_x_forwarded_for&&ref=$http_referer&&uct=$upstream_connect_time&&uht=$upstream_header_time&&urt=$upstream_response_time&&ucs=$upstream_cache_status&&h3=$http3';
```

DiegoRBaquero: so the http3 status can be parsed and sent to the log ingestor

container/nginx/confs/proxy.tlsconf
joaosa: Alright let's go with that just in case

DiegoRBaquero (approved):

container/nginx/confs/proxy.tlsconf
```diff
@@ -37,6 +38,12 @@ server {
 
     resolver    8.8.8.8 1.1.1.1 valid=600s;
 
+    # Add Alt-Svc header to negotiate HTTP/3.
+    add_header alt-svc 'h3-27=":443"; ma=86400, h3-28=":443"; ma=86400, h3-29=":443"; ma=86400, h3=":443"; ma=86400';
```
DiegoRBaquero, line 42: The add_header directives will be lost. Sadly they need to be duplicated wherever we want to return them

> [There could be several add_header directives. These directives are inherited from the previous configuration level if and only if there are no add_header directives defined on the current level.](http://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header)

DiegoRBaquero (changes requested):

container/nginx/confs/proxy.tlsconf
joaosa: Alright I see `There could be several add_header directives. These directives are inherited from the previous configuration level if and only if there are no add_header directives defined on the current level.`

DiegoRBaquero (approved):

PR #65, opened 2022-09-16
Upgrade to http3 on all responses

joaosa:
- Given receiving alt-svc is required to upgrade to h3, let's try to allow clients to do it on all possible responses.

### DIFF
--- a/container/nginx/confs/proxy.tlsconf
+++ b/container/nginx/confs/proxy.tlsconf
@@ -18,6 +18,8 @@ map $format_query $format_final {
     "raw"   "raw";
 }
 
+# The Alt-Svc header is used to negotiate HTTP/3
+# QUIC-Status is used to report HTTP3 usage back to the client
 server {
     listen 443 default_server ssl http2;
     listen 443 http3 reuseport;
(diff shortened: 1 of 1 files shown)


DiegoRBaquero (approved):
I wish they had a flag to just have it inherited by downwards blocks, ffs

gruns:
> I wish they had a flag to just have it inherited by downwards blocks, ffs

they do, no? `add_header` works in `server` blocks:

  * https://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header

so

```
add_header Alt-Svc 'h3-27=":443"; ma=86400, h3-28=":443"; ma=86400, h3-29=":443"; ma=86400, h3=":443"; ma=86400';
add_header QUIC-Status $http3;
```

can be in the `server` directive and then those headers would be returned in every response (i think; quick to double check)

DiegoRBaquero:
It doesn’t fall into child blocks, sadlyOn 16/09/2022, at 4:10 PM, Ansgar Grunseid ***@***.***> wrote:﻿

I wish they had a flag to just have it inherited by downwards blocks, ffs

they do, no? add_header works in server blocks:

https://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header

so
add_header Alt-Svc 'h3-27=":443"; ma=86400, h3-28=":443"; ma=86400, h3-29=":443"; ma=86400, h3=":443"; ma=86400';
add_header QUIC-Status $http3;

can be in the server directive and then those headers would be returned in every response (i think; quick to double check)

PR #73, opened 2022-09-27
Use quiche as the quic implementation

joaosa:
This is an attempt at using a more battle-hardened and actively developed quic implementation. Hopefully this will bring us better TTFB performance.
One drawback is that this brings back BoringSSL and that disables OCSP stapling (we can enable it, but we need to manage the `ssl_stapling_file` manually. More info on how this matters [here](https://markvanlent.dev/2014/04/19/ocsp-stapling-in-nginx/)). Cloudflare has a bunch of things to say about BoringSSL [here](https://blog.cloudflare.com/make-ssl-boring-again/).

My approach was heavily based on https://github.com/patrikjuvonen/docker-nginx-http3/:
- Installed rust so we get the latest cargo and things build as they should (had issues with the stock Debian version);
- Got the two patches: one for OCSP+BoringSSL (AFAICT this enables `ssl_stapling_file`) and the important [one](https://raw.githubusercontent.com/kn007/patch/1062e64ead7e1b21a52392cdd02d1d5bc631d231/nginx_with_quic.patch) which joins nginx and quiche into a happy http3 delivery machine (or so I hope).
- Built nginx with these patches.
- Tweaked the nginx config so it works with quiche.

Maintainability-wise this is a harder path to walk in, as we'll relying on 3rd party patches. This could be, however, a sweet spot until nginx releases official quic support.

Let me know if this makes sense!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,6 +42,7 @@ ARG CONFIG="--prefix=/etc/nginx \
 FROM debian AS build
 
 ARG NGINX_VERSION
+ARG NGINX_BRANCH=quic
 # https://hg.nginx.org/nginx-quic/shortlog/quic
 ARG NGINX_COMMIT=98e94553ae51
 # https://github.com/google/ngx_brotli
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -48,69 +52,63 @@ ARG CONFIG="--prefix=/etc/nginx \
...
+ && git submodule update --init --depth 1
+
+# using boringssl here
+RUN echo "Cloning and getting the quiche patches $QUICHE_COMMIT" \
+ && git clone --recursive https://github.com/cloudflare/quiche \
+ && cd /usr/src/quiche \
+ && git checkout --recurse-submodules $QUICHE_COMMIT \
+ && cd /usr/src \
+ && wget -q https://raw.githubusercontent.com/kn007/patch/1062e64ead7e1b21a52392cdd02d1d5bc631d231/nginx_with_quic.patch \
+ && wget -q https://raw.githubusercontent.com/kn007/patch/cd03b77647c9bf7179acac0125151a0fbb4ac7c8/Enable_BoringSSL_OCSP.patch
```
DiegoRBaquero, line 94: What's this patch for if stapling will be disabled?

DiegoRBaquero (approved):

Dockerfile
joaosa: I added it in case wanted to enable the stapling file. We can remove it if we don't think we'll use that feature

DiegoRBaquero: Gotcha, let's move ahead and see if performance improves. If it does, we can figure out a way to do the manual stapling file

PR #84, opened 2022-10-20
Bump nginx to 1.23.2

joaosa:
There are a few SSL session improvements and a security patch.

See the full changelog [here](http://nginx.org/en/CHANGES).


Dockerfile
```diff
@@ -6,7 +6,7 @@ FROM debian AS build
 ARG NGINX_VERSION
 # https://hg.nginx.org/nginx
 ARG NGINX_BRANCH=default
-ARG NGINX_COMMIT=ba5cf8f73a2d
+ARG NGINX_COMMIT=aa901551a7eb
```
DiegoRBaquero, line 9: Doesn't line 1 need to be updated as well?

DiegoRBaquero (approved):

Issue #82, opened 2022-10-19
add documentation to the readme for node operators


gruns:
add documentation about running a node beyond the l1 setup instructions to the readme

added documentation should include:

* how to create a filecoin wallet if they dont have one already
* how to set up emissary and connect their fil wallet with emissary. screenshots likely helpful
* links and instructions on how to monitor their node, earnings, and stats once set up:
    * https://dashboard.strn.network/
    * https://orchestrator.strn.pl/stats
* any other tips and tricks you can think of
    * @DiegoRBaquero @joaosa how have you guys been monitoring nodes?

DiegoRBaquero:
@guanzo anything missing?

guanzo:
"How to change your FIL address"?

Gracefully stop node. Change env var. Restart node.

PR #85, opened 2022-10-20
Node operator guide

guanzo:
#82

Adds documentation for

* Creating a FIL wallet. Do these choices look okay?
* How to monitor nodes
* How to receive payments

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,9 @@ If you are switching networks, follow these steps:
 
 ### Obtaining a Filecoin wallet address
 
-You need to create a wallet if you don't already have one.
+You need to own a Filecoin wallet to receive FIL payments.
+
+* [Official Filecoin wallet documentation](https://docs.filecoin.io/get-started/overview/#wallets)
 
 * If you have an account on a Centralized Exchange (Coinbase, Binance, etc.) that supports Filecoin,
 go through the steps to deposit Filecoin and you'll be given an wallet address. This is recommended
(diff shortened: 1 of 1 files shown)


DiegoRBaquero (approved):
Are there not better, more frequently updated, guides for the FIL wallet section that are out there in the ecosystem?

guanzo:
The official docs https://docs.filecoin.io/get-started/overview/#wallets
Should we link to that?

DiegoRBaquero:
> The official docs https://docs.filecoin.io/get-started/overview/#wallets Should we link to that?

That works, we don't want to burden ourselves with that section

gruns:
> That works, we don't want to burden ourselves with that section

agreed. we definitely do *not* want to have instructions on how/where to set up a filecoin wallet in our docs. nor do we want to maintain this. we should link elsewhere

that said, the official docs at https://docs.filecoin.io/get-started/overview/#wallets are bad. theyre about wallet implementations

1. i think most node ops who dont already have a fil wallet will want to use a hosted wallet, like coinbase or binance
2. the linked instructions dont even mention such hosted wallet
3. like how many node ops will want to set up and run lotus just to get a fil wallet? my guess: zero

@guanzo lets
1. look for guides on how to get a wallet, including hosted wallets
2. if we cant find any, who should we speak with at PL to write such docs? bc this would be useful for many outside of saturn, too

PR #86, opened 2022-10-21
Make 0-RTT explicit to the backend

joaosa:
tl;dr: TlS 1.3 0-RTT leaves applications vulnerable to replay attacks. There are also issues with forward secrecy, but there is a little we can do there.

One way to counter this is to explicitly signal one's backend that early data is being sent. This way, that backend can decide to drop a given request for non-idempotent operations.

I don't think this is an actual risk since, currently, we're only distributing static content.
This won't matter for a typical `curl https://strn.pl/ipfs/QmQ2r6iMNpky5f1m4cnm3Yqw8VSvjuKpTcK1X7dBR1LkJF/cat.gif`.

It's something important to keep in mind however. More info [here](https://blog.trailofbits.com/2019/03/25/what-application-developers-need-to-know-about-tls-early-data-0rtt/).

### DIFF
--- a/container/nginx/conf.d/register.conf
+++ b/container/nginx/conf.d/register.conf
@@ -7,4 +7,6 @@ location = /register-check {
     proxy_set_header        X-Real-IP       $remote_addr;
     proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header        Connection      "";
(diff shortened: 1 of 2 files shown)


DiegoRBaquero (approved):
We don't really need this until we actually have such endpoints in the shim. Not a security risk right now, but on the side of prevention, what's the perf impact of just adding a dynamic header to proxy back? I am sure it's minimal, but is it negligible?

joaosa:
Since we can't be entirely sure whether there's any measurable performance impact, let's close this for now and apply this when there's a usecase 👍

PR #112, opened 2022-11-18
Update denylist.conf

RobQuistNL:
Block phishing site

Signed-off-by: Rob Quist <RobQuistNL@users.noreply.github.com>


DiegoRBaquero:
Already banned in v479

PR #119, opened 2022-11-22
Integrate IPFS GW requests. 

guanzo:
Gateway requests are coming in as subdomain CIDs

```
https://bafybeibr4hnls245ttz7ukvqmsjouemvfmw3m6zsm7q5kesmok2ynxjova.ipfs.dweb.link
https://bafybeicsuh7zy3mkt2crzwnidg35mwrrzyueruwepeza7ukmoycp6nemei.ipfs.dweb.link/1003.png
```

The L1 routing and logging logic only handles path CIDs. The easiest solution I found is to rewrite the URLs as path CIDs and send them to our existing handler with `proxy_pass`. 

```
http://127.0.0.1/ipfs/bafybeibr4hnls245ttz7ukvqmsjouemvfmw3m6zsm7q5kesmok2ynxjova
http://127.0.0.1/ipfs/bafybeicsuh7zy3mkt2crzwnidg35mwrrzyueruwepeza7ukmoycp6nemei/1003.png
```

This way we reuse everything: the cache (important), the routing, the logging, etc. At the cost of an extra, localhost http request. Seems okay?

In order to do the handoff, I created a new server block that listens on port 80 for the hostname `handoff.strn.localhost`. It feels kinda hacky but it works pretty nicely. 

* We don't need to listen to a new port beyond 80/443. 
* It's not a real domain name so no external requests can hit this server, only requests from the adjacent server block can. 
* The `handoff.strn.localhost` host name will be logged. This lets us easily identify IPFS GW requests, previously there wasn't an explicit field for that.
  *  Any opinions on this host name? 
* It's simple.

Does this strategy look okay? Any other implementation ideas?

For posterity: The gateway is overriding the `core.strn.pl` hostname with the subdomain CID. An example in curl: 
`curl 'https://core.strn.pl/' -H "Host: bafybeibr4hnls245ttz7ukvqmsjouemvfmw3m6zsm7q5kesmok2ynxjova.ipfs.dweb.link"`.
Then the nginx logs will display the url as `https://bafybeibr4hnls245ttz7ukvqmsjouemvfmw3m6zsm7q5kesmok2ynxjova.ipfs.dweb.link`


container/nginx/confs/tls_proxy.conf
```diff
@@ -23,6 +23,41 @@ server {
...
+    ssl_session_tickets         on;
+    ssl_buffer_size             4k;
+    ssl_stapling                on;
+    ssl_stapling_verify         on;
+    ssl_trusted_certificate     /usr/src/app/shared/ssl/node.crt;
+    ssl_early_data              on;
+
+    # endpoint is ipfs/ipns
+    # assumes gateway host is dweb.link. May not be the case in the future.
+    server_name ~^(?<cid>.+)\.(?<endpoint>.+)\.dweb\.link;
```
DiegoRBaquero, line 45: Will this work even if we don't have a wildcard cert for that domain?

container/nginx/confs/non_tls_proxy.conf
```diff
@@ -6,3 +6,14 @@ server {
     include /etc/nginx/conf.d/register.conf;
     include /etc/nginx/conf.d/shared.conf;
 }
+
+server {
+    listen 80;
```
DiegoRBaquero, line 11: Will never arrive here: https://github.com/protocol/bifrost-infra/commit/d700bfeba3a26be72c7ef2ebc864e2c0635ceb2c#diff-a114c9e56a17ead31a0d56c0894354aa74791166e6bc1ca44df5d19a55b70888R159

container/nginx/confs/tls_proxy.conf
DiegoRBaquero: GW must be ignoring cert validity

container/nginx/confs/non_tls_proxy.conf
guanzo: Ya this is just for local dev, when L1 is in non-TLS mode.

container/nginx/confs/tls_proxy.conf
guanzo: `server_name`  just checks `Host` header, TLS shouldn't be involved. (I think)

DiegoRBaquero (approved):

Issue #115, opened 2022-11-19
nginx: [emerg] chown("/usr/src/app/shared/nginx_cache", 101) failed (1: Operation not permitted)


vorburger:
I am exploring running the `ghcr.io/filecoin-saturn/l1-node:main` container in another container runtime environment (k8s) than on Docker via your [`run.sh`](https://github.com/filecoin-saturn/L1-node/blob/main/run.sh) and noticed the following error:

    2022/11/19 23:42:19 [emerg] 29#29: chown("/usr/src/app/shared/nginx_cache", 101) failed (1: Operation not permitted)
    nginx: [emerg] chown("/usr/src/app/shared/nginx_cache", 101) failed (1: Operation not permitted)

I suspect that this is because on (my) k8s the file permission on the `/usr/src/app/shared/` _volume_ is slightly different than on Docker?

I'm not sure you if this is a real problem or can just be ignored.

vorburger:
> I'm not sure you if this is a real problem or can just be ignored.

This log message is a "real" critical (note `emerg` = Nginx "Emergency") problem which blocks orderly start-up. 

#120 raised to make this easier to understand for future users.

> I suspect that this is because on (my) k8s the file permission (...)

It's actually not really that, the reason was just that I tried to run it with more limited Linux capabilities than `root`, for a better security posture. This would be similar e.g. to "[rootless](https://docs.docker.com/engine/security/rootless/)" Docker (or [Podman](https://podman.io/)), where I suspect one woul also run int this error (not yet tested).

vorburger:
https://github.com/docker-library/docs/tree/master/nginx#running-nginx-as-a-non-root-user 😹

PR #120, opened 2022-11-23
Make start.sh abort and fail if one of its steps failed

vorburger:
This seems like a good idea, and could reduce confusion?

One example, among many different possible failures obviously, is what I ran into in #115: `nginx` failed to start, but the shim ran anyway. That's confusing, and it's probably safer to "fail fast" and abort on an `emerg` from Nginx, or any similar launch failures.

Please note that I HAVE NOT TESTED THIS 😈 at all... Does this project run integration tests on PRs?

Would you need me to rebuild and run with this to make sure it's fine? Or can you test it?

### DIFF
--- a/container/shim/start.sh
+++ b/container/shim/start.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -eo pipefail
 
 # patch DNS to use the ones the host passed to docker
 # we grab the top two, so we don't potentially load balance over a ton of resolvers
--- a/container/shim/start.sh
+++ b/container/shim/start.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -eo pipefail
+set -eou pipefail
 
 # patch DNS to use the ones the host passed to docker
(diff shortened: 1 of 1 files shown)


joaosa:
> This seems like a good idea, and could reduce confusion?
> 
> One example, among many different possible failures obviously, is what I ran into in #115: `nginx` failed to start, but the shim ran anyway. That's confusing, and it's probably safer to "fail fast" and abort on an `emerg` from Nginx, or any similar launch failures.
> 
> Please note that I HAVE NOT TESTED THIS 😈 at all... Does this project run integration tests on PRs?
> 
> Would you need me to rebuild and run with this to make sure it's fine? Or can you test it?

Thanks @vorburger! Your approach makes sense and I would even extend it to `set -euo pipefail` 👍 

I'll try to test this when I get a chance. This project does run integration tests, but they do not cover all aspects of the codebase :)

vorburger:
> I would even extend it to `set -euo pipefail` +1

done!

DiegoRBaquero (approved):

PR #122, opened 2022-11-23
docs: Clarify whether a fixed static IP from ISP is required on README

vorburger:
@DiegoRBaquero as per chat - would something like this make sense? Do "suggest changes" (with GitHub ```suggestion thing!) if you see any way to state more clearly what is recommend and what's actually required, and perhaps what "penality" (?) if any there is.


vorburger:
Closing in favour of 6acf5a22f8bd100bd47dc0eaa179ab1c4e0fb20a!

PR #127, opened 2022-11-26
docs: Elaborate on the Saturn volume content

vorburger:
Am I understanding this correctly, and would it be worthwhile to clarify this, in the README?

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,8 @@ git commit -m "my commit message [skip ci]"
 
 ## Files
 
+### In git
+
 #### nginx configuration
 
 `nginx/` contains the nginx configuration of the caching proxy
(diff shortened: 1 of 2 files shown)


DiegoRBaquero:
This is great documentation, however it should not be in the instructions

vorburger:
> This is great documentation, however it should not be in the instructions

ACK. How about I `mv` this into... say.. `docs/architecture/storage.md` ? (Proposing the `/architecture/` as a humble starting point for more future "how does it work internally" documentation, clearly separated from "how do I just run it" type documentation.) Or just `docs/configuration.md`? Or do you simply want it in that FAQ doc?

vorburger:
> `architecture` would be great

OK, I've now moved it out of the `README` - better like this?

DiegoRBaquero (approved):

PR #135, opened 2022-12-03
Add docker.io/library/ prefix to FROM in Dockerfile

vorburger:
Because there is a world of container registries
beyond ;) good ol' Docker Hub, and it's much better
and clearer to specify WHERE one wants to grab
base images FROM.

This also has the nice side effect :) of working around
https://github.com/containers/podman/issues/16629
for me personally (but it's typically a Good Idea
to do this regardless of that).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 ARG NGINX_VERSION="1.23.2"
 ARG NGINX_NAME="nginx-${NGINX_VERSION}"
 
-FROM debian AS build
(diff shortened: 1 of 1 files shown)


DiegoRBaquero:
Where is this URL documented?

vorburger:
> Where is this URL documented?

This is very common, [see all of these gazillions of other places using this](https://github.com/search?q=docker.io%2Flibrary&type=code).

Would you merge it, if I fixed the conflict?

PR #137, opened 2022-12-03
Allow custom HTTP & HTTPS port in run.sh

vorburger:
Because some people like developing where
"rootlessport cannot expose privileged port 80".
(One possible alternative with podman or rootless Docker, is
"'net.ipv4.ip_unprivileged_port_start=80' to /etc/sysctl.conf",
but adding this is simpler and permits running:

    HTTP_PORT=8080 HTTPS_PORT=8443 scripts/run.sh

### DIFF
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -4,6 +4,8 @@
 : "${NODE_OPERATOR_EMAIL:=dev@strn.pl}"
 : "${IPFS_GATEWAY_ORIGIN:=https://ipfs.io}"
 : "${ORCHESTRATOR_REGISTRATION:=true}"
+: "${HTTP_PORT:=80}"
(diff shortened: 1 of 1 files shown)


vorburger:
> Orchestrator won't accept other ports, so check would fail

It doesn't 😄  when used like this, which is what I am currently using to contribute:

    ORCHESTRATOR_REGISTRATION=false HTTP_PORT=8080 HTTPS_PORT=8443 scripts/run.sh

Without this, it seems to be impossible to run it to test contributions to the project if one is also running a Saturn Node at home (unless one has 2 separate fixed IPs, which is not very common). Unless there is a better way which we could document?

DiegoRBaquero (approved):

Issue #124, opened 2022-11-26
Add a HEALTHCHECK to the container


vorburger:
I suspect that if the container had a [`HEALTHCHECK`](https://docs.docker.com/engine/reference/builder/#healthcheck) this will make what #32 #109 #123 are after easier. Something could also be used to [Configure Liveness, Readiness and Startup Probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) for running this container in Kubernetes (k8s). Possibly also useful as [Lifecycle hooks](https://containrrr.dev/watchtower/lifecycle-hooks/) with Watchtower.

It's basically just an executable (binary, script, whatever) in the container image that can test if its working (not faulty).

DiegoRBaquero:
I didn't know about this, this is great! Could be used to test the test CID locally

vorburger:
> I didn't know about this, this is great! 

You're very welcome! (Do contributors to this project also get FIL for friendly good tips? Just joking!)

> Could be used to test the test CID locally

What is that? [According](https://github.com/filecoin-saturn/L1-node/search?q=test+cid) to a quick [read the source, Luke](https://blog.codinghorror.com/learn-to-read-the-source-luke/), I suspect that's `QmXjYBY478Cno4jzdCcPy4NcJYFrwHZ51xaCP8vUwN9MGm`... I have suggested that on PR #142 from @holdenk.

Issue #114, opened 2022-11-19
Container should run with docker run --read-only and not write to its root file system (anywhere else than the mounted volume)


vorburger:
Looking at [`run.sh`](https://github.com/filecoin-saturn/L1-node/blob/main/run.sh) made me think that the `ghcr.io/filecoin-saturn/l1-node:main` container only needs to be able to write `/usr/src/app/shared`, but this does not appear to be so; if I run this image with everything but this read-only (e.g. using `readOnlyRootFilesystem: true` in Kubernetes; but this isn't K8s specific per se, you could see the same with simple Docker) there are a number of errors:

```
sed: couldn't open temporary file /etc/nginx/confs/sedRfbydb: Read-only file system
Sat Nov 19 23:02:42 UTC 2022 [container] booting
Sat Nov 19 23:02:42 UTC 2022 [container] CPUs: 12
Sat Nov 19 23:02:42 UTC 2022 [container] Memory: MemTotal:65411432kB MemFree:62972952kB MemAvailable:63891360kB
Sat Nov 19 23:02:42 UTC 2022 [container] Disk: /dev/nvme1n1p1  954G  6.7G  947G   1% /usr/src/app/shared
mkdir: cannot create directory '/usr/src/app/shared/ssl': Permission denied
Sat Nov 19 23:02:42 UTC 2022 [container] SSL config unavailable, starting non-TLS nginx and node shim
cp: cannot create regular file '/etc/nginx/conf.d/L1.conf': Read-only file system
2022/11/19 23:02:42 [warn] 29#29: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:4
nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:4
2022/11/19 23:02:42 [warn] 29#29: duplicate MIME type "text/html" in /etc/nginx/nginx.conf:66
nginx: [warn] duplicate MIME type "text/html" in /etc/nginx/nginx.conf:66
2022/11/19 23:02:42 [emerg] 29#29: open() "/etc/nginx/conf.d/L1.conf" failed (2: No such file or directory) in /etc/nginx/conf.d/proxy.conf:11
nginx: [emerg] open() "/etc/nginx/conf.d/L1.conf" failed (2: No such file or directory) in /etc/nginx/conf.d/proxy.conf:11
[...]

PR #142, opened 2022-12-03
Add a healthcheck.

holdenk:
See https://github.com/filecoin-saturn/L1-node/issues/124
We can expand on the healthcheck to do more than just check the localhost HTTP but figured this was a good start.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,9 @@ ARG NGINX_NAME="nginx-${NGINX_VERSION}"
 
 FROM debian AS build
 
+HEALTHCHECK --interval=30s --timeout=3s --start-period=60s \
+  CMD (curl -f http://localhost/) || exit 1
+
 ARG NGINX_VERSION
 # https://hg.nginx.org/nginx
 ARG NGINX_BRANCH=default


Dockerfile
```diff
@@ -3,6 +3,9 @@ ARG NGINX_NAME="nginx-${NGINX_VERSION}"
 
 FROM debian AS build
 
+HEALTHCHECK --interval=30s --timeout=3s --start-period=60s \
+  CMD (curl -f http://localhost/) || exit 1
```
vorburger, line 7: HTTP GET / responds with a 302 Moved Temporarily to Location: https://strn.network - which responds with 403. I think this is... "suboptimal", for several reasons:

A. Healthcheck on a URL which redirects, and externally at that not just internally, isn't clearly defined. E.g. on K8s using `GET /` as a [Liveness, Readiness and/or Startup Probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) leads to [a confusing weird `ProbeWarning`](https://github.com/kubernetes/website/issues/30849) - I just ran into it, with this container... 🥲  (I can't remember whether a container's `HEALTHCHECK` automagically become a probe on K8s, but even if it doesn't, another check would be be better, IMHO.)

B. It's not clear to me yet if just Nginx responding on / really means the container is actually fully ready.

@DiegoRBaquero in #124 mentions _"Could be used to test the test CID locally"_ and I just dug a bit more into what he may have meant by that... how about we instead propose using this:

```suggestion
  CMD (curl -f http://localhost/ipfs/QmXjYBY478Cno4jzdCcPy4NcJYFrwHZ51xaCP8vUwN9MGm/) || exit 1
```

vorburger (changes requested):

Dockerfile
```diff
@@ -3,6 +3,9 @@ ARG NGINX_NAME="nginx-${NGINX_VERSION}"
 
 FROM debian AS build
 
+HEALTHCHECK --interval=30s --timeout=3s --start-period=60s \
```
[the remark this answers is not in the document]
holdenk: Ah gotcha, yeah a halth check in the build container is probably not very useful.

vorburger (approved):
LGTM! (But I haven't tested it in production.)

Dockerfile
```diff
@@ -85,6 +85,9 @@ RUN echo "Cloning nginx and building $NGINX_VERSION (rev $NGINX_COMMIT from '$NG
 
 FROM nginx:${NGINX_VERSION}
 
+HEALTHCHECK --interval=30s --timeout=3s --start-period=60s \
+  CMD (curl -f http://localhost/ipfs/QmXjYBY478Cno4jzdCcPy4NcJYFrwHZ51xaCP8vUwN9MGm/) || exit 1
```
vorburger, line 89: Wait, does this actually work? If I run `curl -f http://192.168.1.xxx:31080/ipfs/QmXjYBY478Cno4jzdCcPy4NcJYFrwHZ51xaCP8vUwN9MGm/` from CLI, it returns 23, see https://everything.curl.dev/usingcurl/returns... which would fail the HEALTCHECK, I think.

vorburger (changes requested):

Dockerfile
holdenk: Weird when I run this inside of the container what I see is:

```
root@DEN:/usr/src/app# curl -f http://localhost/ipfs/QmXjYBY478Cno4jzdCcPy4NcJYFrwHZ51xaCP8vUwN9MGm/
<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>
root@DEN:/usr/src/app# echo $?
0

```

vorburger: Duh, yeah, please ignore that... sorry. But while looking into this, I've noticed that as-is it would actually pass (exit code 0) whether or not it "really worked" (e.g. even if it's a  wrong IPFS CID, or there is a problem to resolve it, it' still always just 301). I've had a closer look at `curl` options, what do you think about we do the following:

    curl --fail --silent --output /dev/null --location --insecure http://178.199.229.8/ipfs/QmXjYBY478Cno4jzdCcPy4NcJYFrwHZ51xaCP8vUwN9MGm

The `--fail` is (your) `-f` just perhaps more explicit in the long form.

The `--silent --output /dev/null` are with an eye towards #114.

The `--location --insecure` follows that 301 redirect - and lets use "test" that it "really worked".

vorburger:
```suggestion
  CMD (curl --fail --silent --output /dev/null --location --insecure http://localhost/ipfs/QmXjYBY478Cno4jzdCcPy4NcJYFrwHZ51xaCP8vUwN9MGm/) || exit 1
```

@holdenk LGTY?

vorburger: @holdenk do you want to click Accept / Apply for this?

PR #168, opened 2023-01-11
Migrating away from update.sh and run.sh

### LINKED ISSUE #109: Replace update.sh with an external dependency
The less code we have to maintain, the less complexity we have to focus on and the better it will be on the long run.

This principle can be applied to `update.sh` which can be replaced by [watchtower](https://github.com/containrrr/watchtower), for instance.

Since we can't have all nodes rebooting at the same time, as that would kill the network, we can use [lifecycle hooks](https://containrrr.dev/watchtower/lifecycle-hooks/) to wait a random time, as it is currently done [here](https://github.com/filecoin-saturn/L1-node/blob/main/update.sh#L49).

[This guide](https://containrrr.dev/watchtower/usage-overview/) offers a good way to get started. [This](https://www.digitalocean.com/community/tutorials/how-to-automatically-update-docker-container-images-with-watchtower-on-ubuntu-22-04) is another good ref.

One should note that watchtower [does not support rollbacks](https://github.com/containrrr/watchtower/issues/90). Doing so would be ideal, as per [this issue](https://github.com/filecoin-saturn/L1-node/issues/32).

### LINKED ISSUE #125: Enable container to terminate cleanly
Could it make sense to improve upon the current approach how to [gracefully stop a node a not receive a penalty](https://github.com/filecoin-saturn/L1-node#stopping-a-node) by enabling the container to terminate cleanly?

As-is currently, it keeps running after `SIGTERM` (so you signal it using e.g. [`docker kill`](https://docs.docker.com/engine/reference/commandline/kill/), wait a fixed amount of time, and then terminate it with [`docker stop`](https://docs.docker.com/engine/reference/commandline/stop/) - which will `SIGTERM` again, wait another 10s, and then `SIGKILL` it.

If there was any way for it to figure out and know by itself when it's "done" and then exit, I have a hunch that this, together with #124, could contribute to making #32, #109 and #123 easier and upgrades faster. Then you could simply `docker stop` [`--time=900`](https://github.com/filecoin-saturn/L1-node/blob/5baf03497c2a708c77779c99d6ba0c3b0af9d479/update.sh#L55) but those 15' would then be an upper bound maximum, not a hard-coded fixed duration anymore.

I don't know enough exact details about the architecture yet, but I suspect that this at least one part of solving this includes "stopping / decelining to accept new HTTP requests, but waiting for Nginx to finish serving the ones it currently has going".

AnomalRoil:
This allows us to:
- deprecate the `run.sh` and `update.sh` scripts which are running with sudo rights and are big security liabilities 
- streamline the way environment variables are set using an `.env` file, no more need to tamper with global env variables
- have a nice docker-based setup using docker compose

I did my best at creating a migration script in `update.sh` too, which should allow everyone to upgrade easily.

Improvements:
- Now the docker container `saturn-node` has its `stop_grace_period` and `stop_signal` set by docker itself, no need to kill and sleep anymore 
- The Docker compose command will fail with an error message if the env variables aren't properly set instead of just complaining in the docker container logs.
- The updates are now being handled by a Watchtower container.

Known difference: 
- I could not replicate the "random delay" in the update script with Watchtower, but the watchtower container starts its timer upon start and so it shouldn't be synced for all nodes
- I've set the Watchtower interval to 1 hour since the update script was sleeping long random delays anyway when updating

It also incidentally fixes #109 and fixes #125.


update.sh
```diff
@@ -9,67 +9,63 @@ if pidof -o %PPID -x "update.sh" > /dev/null; then
...
+compose_file=$SATURN_HOME/docker-compose.yml
+env_file=$SATURN_HOME/.env
+
+echo "$(date -u) Starting migration to docker compose setup."
+echo "The auto-update script was deprecated to migrate to a docker compose setup. Updates are handled by a Watchtower container now."
+echo "The run script was deprecated too. To start or stop your node use 'docker compose -f $compose_file up -d' or 'docker compose -f $compose_file down' respectively."
+
+if [ -f "$compose_file" ]; then
+    echo "We have migrated to a docker compose setup to reduce risks caused by these update and run scripts. You can delete these scripts and use the Docker Compose workflow now."
+else
```
AnomalRoil, line 21: This is meant to prevent (security) issues where leftover cron jobs might execute this script every 5 minutes.
After running this update.sh once, if it worked it won't run anymore, not downloading nor running any random things with `sudo` anymore.

docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+services:
+  saturn-node:
+    image: ghcr.io/filecoin-saturn/l1-node:${SATURN_NETWORK:-test}
+    container_name: saturn-node
+    restart: always
+    environment:
+      SATURN_NETWORK: "${SATURN_NETWORK:?please make sure to set your SATURN_NETWORK environment variable with either 'test' or 'main' in the .env file}"
+      FIL_WALLET_ADDRESS: "${FIL_WALLET_ADDRESS:?please make sure to set your FIL_WALLET_ADDRESS environment variable in the .env file}"
+      NODE_OPERATOR_EMAIL: "${NODE_OPERATOR_EMAIL:?please make sure to set your NODE_OPERATOR_EMAIL environment variable in the .env file}"
+      SPEEDTEST_SERVER_CONFIG: "${SPEEDTEST_SERVER_CONFIG}"
```
AnomalRoil, line 11: I could have used only the `.env` file, but I figured people might be taking the docker-compose file as an example to craft their own, and so better include these here too.

joaosa: that seems reasonable to me 👍

docker-compose.yml
```diff
@@ -0,0 +1,42 @@
+version: "3.8"
+services:
+  saturn-node:
+    image: ghcr.io/filecoin-saturn/l1-node:${SATURN_NETWORK:-test}
+    container_name: saturn-node
+    restart: always
```
joaosa, line 6: @DiegoRBaquero do we want to default to `unless-stopped` instead?

update.sh
```diff
@@ -9,67 +9,63 @@ if pidof -o %PPID -x "update.sh" > /dev/null; then
...
-  sudo docker image prune -f
+  echo "Draining $SATURN_NETWORK network L1 node... "
+  sudo docker stop --time 900 saturn-node || true
 
-  echo "Updated to latest version successfully!"
-else
-  echo "Saturn $SATURN_NETWORK network L1 node up to date"
+  echo "Restarting with docker compose now..."
+  sudo docker rm -f saturn-node || true
+  sudo -E docker compose -f $compose_file pull
```
joaosa, line 62: @AnomalRoil just to double-check: is the reason for this second pull to ensure we're running fresh images post-drain?

joaosa:
> This allows us to:
> 
>     * deprecate the `run.sh` and `update.sh` scripts which are running with sudo rights and are big security liabilities
> 
>     * streamline the way environment variables are set using an `.env` file, no more need to tamper with global env variables
> 
>     * have a nice docker-based setup using docker compose
> 
> 
> I did my best at creating a migration script in `update.sh` too, which should allow everyone to upgrade easily.
> 
> Improvements:
> 
>     * Now the docker container `saturn-node` has its `stop_grace_period` and `stop_signal` set by docker itself, no need to kill and sleep anymore
> 
>     * The Docker compose command will fail with an error message if the env variables aren't properly set instead of just complaining in the docker container logs.
> 
>     * The updates are now being handled by a Watchtower container.
> 
> 
> Known difference:
> 
>     * I could not replicate the "random delay" in the update script with Watchtower, but the watchtower container starts its timer upon start and so it shouldn't be synced for all nodes
> 
>     * I've set the Watchtower interval to 1 hour since the update script was sleeping long random delays anyway when updating
> 
> 
> It also incidentally fixes #109 and fixes #125.

@AnomalRoil this seems like a solid approach. I'm particularly glad you tackled #109 and improved the env var setup 💪. I'm a bit more concerned about the "random delay" part, but I'll be testing this out. Thanks a lot!

docker-compose.yml
DiegoRBaquero: Yes

docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+    environment:
+      SATURN_NETWORK: "${SATURN_NETWORK:?please make sure to set your SATURN_NETWORK environment variable with either 'test' or 'main' in the .env file}"
+      FIL_WALLET_ADDRESS: "${FIL_WALLET_ADDRESS:?please make sure to set your FIL_WALLET_ADDRESS environment variable in the .env file}"
+      NODE_OPERATOR_EMAIL: "${NODE_OPERATOR_EMAIL:?please make sure to set your NODE_OPERATOR_EMAIL environment variable in the .env file}"
+      SPEEDTEST_SERVER_CONFIG: "${SPEEDTEST_SERVER_CONFIG}"
+    ulimits:
+        nofile:
+          soft: 65536
+          hard: 65536
```
DiegoRBaquero, line 15: This needs to match current value of one million

docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+    ports:
+      - 80:80
+      - 443:443
+    network_mode: host
+    stop_signal: SIGTERM
```
DiegoRBaquero, line 20: Does this override the Dockerfile one? (`SIGQUIT`)?

This (`SIGTERM`) is our drain signal

DiegoRBaquero: Yes it does

DiegoRBaquero:
This is great, thank you for your contribution. I think it'll be ready to go once requested changes are tackled

.env
```diff
@@ -0,0 +1,4 @@
+SATURN_NETWORK="test"
+FIL_WALLET_ADDRESS=""
+NODE_OPERATOR_EMAIL=""
+SPEEDTEST_SERVER_CONFIG=""
```
YurkoWasHere, line 4:
```suggestion
SPEEDTEST_SERVER_CONFIG=""
SATURN_HOME=""
```

Missing the saturn_home variable

update.sh
AnomalRoil: Yes, because the `--time 900` in `docker stop` means it can take up to 15 minutes to shut down.

docker-compose.yml
AnomalRoil: FYI, on Amazon Linux 65536 is the max value, IIRC. But done.

DiegoRBaquero (approved):

Issue #111, opened 2022-11-17
Map docker images to their git tags


joaosa:
Versioning is displayed per node on the stats page ([production](https://orchestrator.strn.pl/stats) and [staging](https://orchestrator.strn-test.pl/stats)). 

This number (e.g. `v475`) corresponds to a CI build number and not to the actual version associated to the git commit which triggered it (e.g. `1668199595`). 
If multiple builds were to occur for a particular commit and those were to produce different docker image artifacts, we wouldn't be able to relate them. One way to fix this is to keep both values in mind and to display them together (e.g. `v1668199595-b475`).

Another aspect to consider is the versioning of docker images. We're currently using mutable docker tags for client convenience. It's easy to ensure we have the latest version with a `docker pull ghcr.io/filecoin-saturn/l1-node:main`. We get environment scoping for free, i.e. the latest staging image does not get mixed with the latest production one.

This is, however, not ideal if we were to need to troubleshoot what is going on with a particular version. As docker tags are mutable, they move to the latest image pushed to that tag. This leaves behind a set of images we have little information about ([see here](https://github.com/filecoin-saturn/L1-node/pkgs/container/l1-node/versions)). The need to debug past images may not be aligned with keeping the network running the latest version, but it could be invaluable for ensuring there is no performance degradation between versions (for instance).

One approach would be to use a single immutable docker image tag (note that `ghrc.io` does not seem to enforce this principle [the way AWS does](https://docs.aws.amazon.com/AmazonECR/latest/userguide/image-tag-mutability.html)) as opposed to rolling tags (i.e. tags which point to the last image pushed with that tag). Following this would ensure a particular image tag cannot be overwritten and will always map to the version it originated from.

This also means giving up the client-side UX of fetching the latest image with a simple docker command. This could be fine if properly documented given version numbers are typically monotonic increasing in their sequence (i.e. version `1668199595` is ensured to come after `1667316793`).
Also noteworthy is the fact this approach doesn't work with [watchtower](https://github.com/containrrr/watchtower/issues/460) for [auto-updates](##Auto-update).

To work around this, we could tag images twice. Both with their git commit tag and target environment. This way we would be able to map one to a particular git commit while preserving the client UX side of things.

vorburger:
> To work around this, we could tag images twice. Both with their git commit tag and target environment. 

+1 - I think this would be a great! FYI I've also seen other container images do exactly this in the past, it's common.

[...]

PR #173, opened 2023-01-17
Tag images with the commit sha and build number

joaosa:
Refer to https://github.com/filecoin-saturn/L1-node/issues/111 for the rationale.

tl;dr this is good for traceability and regression testing.

I went with `${commit_sha}-${build_number}` as the tag format. lmk if that makes sense.


.github/workflows/release.yml
```diff
@@ -98,7 +98,9 @@ jobs:
         with:
           context: .
           push: true
-          tags: ghcr.io/filecoin-saturn/l1-node:${{ env.SATURN_NETWORK }}
+          tags: |
+            ghcr.io/filecoin-saturn/l1-node:${{ env.SATURN_NETWORK }}
+            ghcr.io/filecoin-saturn/l1-node:${{ github.sha }}-${{ github.run_number}}
```
DiegoRBaquero, line 103:
```suggestion
            ghcr.io/filecoin-saturn/l1-node:${{ github.run_number}}_${{ github.sha }}
```

DiegoRBaquero: This would match what's in code and is shared to the orch

DiegoRBaquero: And would be better to have the short sha to exactly match: https://stackoverflow.com/questions/59810838/how-to-get-the-short-sha-for-the-github-workflow

DiegoRBaquero (approved):

PR #175, opened 2023-01-17
use logfmt instead of custom log format

kwypchlo:
- use logfmt instead of custom log format
  - logfmt is supported out of the box and almost fastest way of parsing logs in [Loki](https://grafana.com/oss/loki/)
- added some more stats for future perf measurements

✅ tested on test environment


DiegoRBaquero (changes requested):
Looks good but seems like the style for functions is being changed, probably prettier? Not in the scope of the PR if it's not being adopted

kwypchlo:
> Looks good but seems like the style for functions is being changed, probably prettier? Not in the scope of the PR if it's not being adopted

Yeah it seems so, I parsed it with eslint to fix the styling.

DiegoRBaquero (approved):

PR #179, opened 2023-01-20
move nginx logs to mounted volume

kwypchlo:
- move nginx log from `/var/log/nginx` to mounted volume `/usr/src/app/shared/nginx_log`
- path consistent with nginx cache location which is `/usr/src/app/shared/nginx_cache`
- because currently nginx logs are not persisted when container is recreated we do not need to worry about past logs

_this is prerequisite for persisting nginx logs and network monitoring that will come in another pull request_


joaosa:
This looks good to me 👍 . Let's see what @DiegoRBaquero has to say :)

DiegoRBaquero (approved):
This is great and will help prevent logs loss!!

PR #182, opened 2023-01-22
correct docker compose to docker-compose

YurkoWasHere:
Update script is currently broken. This patch attempts to fix it.

There should also be a migration path for people that run docker to docker-compose

Fixes:
- `docker compose` should be `docker-compose`
- `docker-compose.yml`  requires escaped $ in watchtower line
- Corrected prettier yml due to CI complaining about it

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -90,7 +90,7 @@ jobs:
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v2
         with:
-          platforms: 'arm64'
+          platforms: "arm64"
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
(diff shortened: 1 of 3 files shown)


YurkoWasHere:
In `main` the .env file pulled form the repo containing blank entries. I think this is a bad UX:

I would suggest

- Migrate current setting to `.env`
https://github.com/filecoin-saturn/L1-node/pull/182/files#diff-e14594f6bb9b29ac0146c04e78ae33e8e92f2f00709fefab97359dea31b94d11L31-L38

- Don't create the .env file at all and require user to manually migrate

YurkoWasHere:
Just came to my attention that "docker compose" is the "new way" of doing things. Perhaps it is an older docker version that I have but in the channel a lot of people are having the same issue.

I will have to investigate this claim.

juliangruber:
> Just came to my attention that "docker compose" is the "new way" of doing things. Perhaps it is an older docker version that I have but in the channel a lot of people are having the same issue.

This is correct, upgrade docker and then `docker compose` will work

PR #184, opened 2023-01-23
Always send updated node id

### LINKED ISSUE #180: Always send most current node ID
Instead of sending the node id from cache. Always inject it using latest node id.

This allows to copy cache folder, or sync them between multiple nodes.

DiegoRBaquero:
Closes #180


container/nginx/conf.d/shared.conf
```diff
@@ -56,14 +56,17 @@ location / {
     proxy_cache_bypass          $arg_nocache;
     proxy_no_cache              $arg_nocache;
 
-    add_header 'Saturn-Cache-Status'            $upstream_cache_status always;
+    if ($node_id) {
+        add_header 'Saturn-Node-Id'                 $node_id always;
+    }
```
kwypchlo, line 61: have you tested that this will not cause any issues?
https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/
http://agentzh.blogspot.com/2011/03/how-nginx-location-if-works.html

container/shim/src/config.js
```diff
@@ -66,12 +67,14 @@ function pVersion(version) {
...
+  let nodeId;
   try {
-    const nodeId = await readFile(NODE_ID_FILE_PATH, "utf8");
-    return nodeId.trim();
+    nodeId = (await readFile(NODE_ID_FILE_PATH, "utf8")).trim();
   } catch (err) {
-    return createNodeId();
+    nodeId = createNodeId();
   }
+  writeFile(NODE_ID_NGINX_CONF, `set $node_id "${nodeId}";`).catch(console.error); // eslint-disable-line no-console
```
kwypchlo, line 76: the first time this file is created, nginx will be already running so this file will not be included and nginx will need to restart or reload config right ?

container/shim/start.sh
```diff
@@ -18,6 +18,7 @@ echo "$(date -u) [container] Disk: $(df -h /usr/src/app/shared | awk '(NR>1)')"
 
 # Create if not exists
 mkdir -p /usr/src/app/shared/ssl
+mkdir -p /usr/src/app/shared/nginx
```
kwypchlo, line 21: can we use nginx_conf here? we already have nginx_cache and nginx_log dirs so it would match the convention 

```suggestion
mkdir -p /usr/src/app/shared/nginx_conf
```

container/nginx/conf.d/shared.conf
DiegoRBaquero: ah, it won't work like this, thing is, it won't be available on first run 🤔

container/shim/src/config.js
DiegoRBaquero: correct, that's why the `if` was there :/

container/shim/start.sh
DiegoRBaquero: I was hoping to eventually have everything inside `/nginx`, logs, ssl, config

container/nginx/conf.d/shared.conf
guanzo: On first run, nginx crashes with 
```
2023/01/24 15:52:01 [emerg] 34#34: unknown "node_id" variable
nginx: [emerg] unknown "node_id" variable
```

but! The container keeps running. The container should exit if nginx fails to start, since its useless if nginx is down.

container/nginx/confs/non_tls_proxy.conf
```diff
@@ -3,6 +3,7 @@ server {
 
     server_name _;
 
+    include /usr/src/app/shared/nginx_conf/*.conf;
```
DiegoRBaquero, line 6: Ah, right, variable is used, and needed

kwypchlo (approved):

joaosa (approved):
LGTM

PR #187, opened 2023-01-24
resumable log ingestor

kwypchlo:
### Issues with current log ingestor:
- truncates the log file on every read leaving file empty every time
  - node operator is left with empty log file and cannot investigate or verify traffic
  - not all logs are submitted (only retrievals are submitted) but all logs are truncated
  - not all metrics from logs are submitted - no past data to backfill missing ones
- susceptible to log line losses
  - queue of logs pending for submission is kept in memory and lost in case of service failure
  - does not retry submitting logs on submission failures
  - truncates file after read while it could have been written to in the meantime

### New log ingestor:
- streams log file in 64 kb chunks to keep memory usage low even when reading multiple GB files
- batches submission to orchestrator into at most 10 MB of payload per single request
- leaves log file intact, does not truncate
- persists byte offset of read and processed logs to resume ( filename with .offsetfile suffix ie. node-access.log.offsetfile )
- continues reading more only once logs were successfully submitted - retries same logs again on submission failure
- persists inode and file size to detect file rotation and reset byte offset
- backwards compatible with current ingestor, drop in replacement

### DIFF
--- a/container/shim/src/bin/shim.js
+++ b/container/shim/src/bin/shim.js
@@ -14,7 +14,7 @@ import {
 } from "../config.js";
 import { trapServer } from "../utils/trap.js";
 import { debug } from "../utils/logging.js";
-import { submitRetrievals, initLogIngestor } from "../modules/log_ingestor.js";
+import startLogIngestor from "../modules/log_ingestor.js";
 import { refreshLocalNodes } from "../modules/local_nodes.js";
 
 debug("Saturn L1 Node");
@@ -43,8 +43,9 @@ setTimeout(async function () {
     });
   }
 
-  // Start log ingestor
-  await initLogIngestor();
+  // run log ingestor process in background (starts immediately and keeps running periodically)
+  startLogIngestor();
+
   refreshLocalNodes();
 }, 500);
 
@@ -59,7 +60,7 @@ trapServer(server);
 
 async function shutdown() {
   try {
-    await Promise.allSettled([submitRetrievals(), ORCHESTRATOR_REGISTRATION ? deregister() : Promise.resolve()]);
+    await Promise.allSettled([startLogIngestor(), ORCHESTRATOR_REGISTRATION ? deregister() : Promise.resolve()]);
   } catch (err) {
     debug(`Failed during shutdown: ${err.name} ${err.message}`);
   } finally {
--- a/container/shim/src/modules/log_ingestor.js
+++ b/container/shim/src/modules/log_ingestor.js
@@ -3,6 +3,7 @@ import fetch from "node-fetch";
 import pLimit from "p-limit";
 import prettyBytes from "pretty-bytes";
 import logfmt from "logfmt";
+import glob from "fast-glob";
 import readlines from "../utils/readlines.js";
 
 import { FIL_WALLET_ADDRESS, LOG_INGESTOR_URL, nodeId, nodeToken, TESTING_CID } from "../config.js";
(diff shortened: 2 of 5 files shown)


container/shim/src/bin/shim.js
```diff
@@ -59,7 +60,7 @@ trapServer(server);
 
 async function shutdown() {
   try {
-    await Promise.allSettled([submitRetrievals(), ORCHESTRATOR_REGISTRATION ? deregister() : Promise.resolve()]);
+    await Promise.allSettled([startLogIngestor(), ORCHESTRATOR_REGISTRATION ? deregister() : Promise.resolve()]);
```
DiegoRBaquero, line 63: The timeout prevents dual interval, but not dual execution, what happens if it's already running when this is called?

container/shim/src/modules/log_ingestor.js
```diff
@@ -41,155 +41,150 @@ const NGINX_LOG_KEYS_MAP = {
...
+ * @returns {Promise<void>} - resolves once the logs are successfully submitted to the orchestrator
+ */
+async function submitLogs(logs) {
+  // cache hit rate for this batch of retrievals
+  const cacheHits = logs.filter(({ cacheHit }) => cacheHit).length;
+  const cacheHitRate = cacheHits / logs.length;
+  const cacheHitRatePercent = Math.round(cacheHitRate * 100);
+
+  // total bytes sent to clients for this batch or retrievals
+  const totalBytesSent = logs.reduce((acc, { numBytesSent }) => acc + numBytesSent, 0);
```
DiegoRBaquero, line 118: logs is being traversed twice here. You can reuse the `reduce` to also count cache hits.

container/shim/src/bin/shim.js
kwypchlo: addressed in https://github.com/filecoin-saturn/L1-node/pull/187/commits/78060a780cee5e8d889400ffd7237779588dfbb4

DiegoRBaquero (approved):

container/shim/src/modules/log_ingestor.js
```diff
@@ -41,155 +43,160 @@ const NGINX_LOG_KEYS_MAP = {
...
-      pending = body.bandwidthLogs.concat(pending);
+        // add parsed log line if it is valid (only valid retrieval)
+        if (parsed) logs.push(parsed);
+      }
+
+      // submit the logs to the log ingestor
+      if (logs.length) {
+        try {
+          // submit parsed logs to the orchestrator
+          await submitLogs(logs);
```
guanzo, line 180: These comments are unnecessary imo. Sometimes code can actually be self documenting. I'm singling out this section but I mean it for the entire PR.

I like comments when they clarify business logic or add additional context. 
I dont like comments when they restate the code, it doesn't add value imo.

So for example instead of 

```js
        // skip empty lines
        if (read.lines[i] === "") continue;
```

you can do
```js
        const isEmptyLine = read.lines[i] === ""
        if (isEmptyLine) continue;
```

and it reads just as well. Except now, if that line of code ever changes, you don't have to update its corresponding comment.

I approve of this PR in its current state, but in the future I'd like to see less comments and more descriptive code.

container/shim/src/modules/log_ingestor.js
```diff
@@ -41,155 +43,160 @@ const NGINX_LOG_KEYS_MAP = {
...
-}
+    if (read.lines.length) {
+      const logs = [];
+
+      for (let i = 0; i < read.lines.length; i++) {
+        // skip empty lines
+        if (read.lines[i] === "") continue;
+
+        // parse the line into an object
+        const parsed = parseSingleLine(read.lines[i]);
```
guanzo, line 170: These kinds of comments are unnecessary imo. Sometimes code can actually be self documenting. I'm singling out this section but I mean it for the whole PR.

I like comments when they clarify business logic or add additional context. 
I dont like comments when they restate the code, it doesn't add value imo.

So for example instead of 

```js
        // skip empty lines
        if (read.lines[i] === "") continue;
```

you can do
```js
        const isEmptyLine = read.lines[i] === ""
        if (isEmptyLine) continue;
```

and it reads just as well. Except now, if that line of code ever changes, you don't have to update its corresponding comment.

I approve of this PR in its current state, but in the future I'd like to see less comments and more descriptive code. Would love to hear others' thoughts on this

guanzo:
lgtm. great job!

container/shim/src/modules/log_ingestor.js
kwypchlo: Good feedback @guanzo! TBH I just got Github Copilot subscription and went a bit crazy with code comments since it was just too easy to generate them. Your example about self documenting code is spot on, I will keep that in mind in the future 💯

PR #206, opened 2023-02-06
initial stub on lassie integration

willscott:
* it looked like the range check was being overly aggressive. okay to scope it down like this?
* thoughts on the proposed header check for when to respond from lassie vs gateway?

### DIFF
--- a/container/shim/src/index.js
+++ b/container/shim/src/index.js
@@ -118,7 +118,7 @@ const handleCID = asyncHandler(async (req, res) => {
 
   debug(`Cache miss for ${req.path}`);
 
-  if (req.headers.origin === "rhea") {
+  if ((req.headers["x-fetcher"] || "").indexOf("bifrost-gateway") > -1) {
     // TODO: respond from lassie
     debug(`Cache miss to serve with lassie`);
   }
--- a/container/shim/src/index.js
+++ b/container/shim/src/index.js
@@ -98,7 +98,7 @@ const handleCID = asyncHandler(async (req, res) => {
 
   res.set("Content-Type", mimeTypes.lookup(req.path) || "application/octet-stream");
 
-  if (req.headers.range) {
+  if (req.headers.range && cid === TESTING_CID) {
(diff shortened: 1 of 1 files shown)


willscott:
> I feel like this is breaking the HTTP spec a bit. 

that was making it harder for some mis-coded browser client to accidentally send that header value. that said, happy to do it as a `x-...` header or similar. what would you prefer for name / value/

guanzo:
> that was making it harder for some mis-coded browser client to accidentally send that header value. 

ya good point. it also makes it harder for well-coded browser clients to intentionally send the header value.  🤔 

ok spit balling here but how about `x-rhea` or `x-fetcher` for the name, and `lassie` as the value. Then if we want to A/B test new lassie versions or something different altogether, we can add more values.

willscott:
`x-fetcher: bifrost-gateway/1.0.0`?

guanzo:
sgtm

guanzo:
Shall we communicate whenever the `bifrost-gateway` version is bumped? Or just do a prefix match.

guanzo (approved):

container/shim/src/index.js
```diff
@@ -118,6 +118,11 @@ const handleCID = asyncHandler(async (req, res) => {
 
   debug(`Cache miss for ${req.path}`);
 
+  if ((req.headers["x-fetcher"] || "").indexOf("bifrost-gateway") > -1) {
```
juliangruber, line 121:
```diff
-   if ((req.headers["x-fetcher"] || "").indexOf("bifrost-gateway") > -1) {
+   if ((req.headers["x-fetcher"] || "").includes("bifrost-gateway")) {
```

Bystander comment: This should be easier to read

willscott: 😅  my JS is a couple years out of date

guanzo:
```js
if (req.headers["x-fetcher"]?.includes("bifrost-gateway"))`
```

prolly the most l33t hacker version

juliangruber: Nice!

PR #207, opened 2023-02-06
Fetcher modules

guanzo:
Refactors code in preparation for lassie. The 3 fetchers (ipfs gateway, L2 node, lassie) are extracted out of index.js and into their own modules.

### DIFF
--- a/container/shim/src/fetchers/ipfs-gateway.js
+++ b/container/shim/src/fetchers/ipfs-gateway.js
@@ -1,0 +1,89 @@
+import { cpus } from "node:os";
+import http from "node:http";
+import https from "node:https";
+
+import { streamCAR } from "./utils/car.js";
+import { proxyRequestHeaders, proxyResponseHeaders, toUtf8 } from "../utils/http.js";
+import { debug } from "../utils/logging.js";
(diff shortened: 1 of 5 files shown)


container/shim/src/fetchers/ipfs-gateway.js
```diff
@@ -0,0 +1,89 @@
...
+import { streamCAR } from "../utils/car.js";
+import { proxyRequestHeaders, proxyResponseHeaders, toUtf8 } from "../utils/http.js";
+import { debug } from "../utils/logging.js";
+import { IPFS_GATEWAY_ORIGIN } from "../config.js";
+
+const GATEWAY_TIMEOUT = 120_000;
+
+const agentOpts = {
+  keepAlive: true,
+  maxSockets: Math.floor(128 / cpus().length),
```
DiegoRBaquero, line 14: For nodes with over 128 CPUs. Also, lower to get rate limited less

```suggestion
  maxSockets: Math.max(Math.floor(100 / cpus().length), 1),
```

DiegoRBaquero (approved):
Ah, I love this extraction from the bloated index.js

PR #209, opened 2023-02-07
Fetch from Lassie

guanzo:
Dependent on https://github.com/filecoin-saturn/L1-node/pull/207, merge that first.

The Lassie daemon is conditionally enabled if the `LASSIE_ORIGIN` env var is set. It's technically overridable but the port is hardcoded to `7766` for the moment.

The L1 will fetch from Lassie if the above env var is set and a request has the header `x-fetcher: bifrost-gateway`

When the L1 receives the CAR file from Lassie, the requested block is returned to the gateway and the request is ended. All subsequent blocks are stored in an **in memory** LRU cache, keyed by CID. Initially the plan was to write the blocks to nginx's cache, but after some research this turned out to be too complex and would delay our timeline. The cache has a max byte size of 1 GB, and node.js max memory has been increased to 2 GB. I figured this was okay because L1's have a minimum RAM requirement of 32GB.


Dockerfile
```diff
@@ -100,6 +100,10 @@ RUN curl -fsSL https://deb.nodesource.com/setup_18.x | bash - \
  jq \
  && rm -rf /var/lib/apt/lists/*
 
+# Download lassie
+RUN curl -L 'https://github.com/filecoin-project/lassie/releases/download/v0.3.1/lassie-v0.3.1-linux-386.tar.gz' -o lassie.tar.gz \
```
joaosa, line 104: Since we support arm64 as well, it would be best to get `https://github.com/filecoin-project/lassie/releases/download/v0.3.1/lassie-v0.3.1-linux-$(uname -m).tar.gz`. To do this you probably need:

```
RUN sh -c 'curl -L "https://github.com/filecoin-project/lassie/releases/download/v0.3.1/lassie-v0.3.1-linux-$(uname -m).tar.gz" -o lassie.tar.gz \
  && tar -C /usr/bin -xzf lassie.tar.gz'
```

joaosa: Or [even better](https://nielscautaerts.xyz/making-dockerfiles-architecture-independent.html) to use `TARGETPLATFORM` and parse that to get the right value depending on the target build.

container/shim/start.sh
```diff
@@ -42,4 +42,15 @@ if [ -n "${IPFS_GATEWAY_ORIGIN:-}" ]; then
...
-exec node src/bin/shim.js
+
+if [ "$LASSIE_ORIGIN" != "" ]; then
+  lassie daemon -p 7766 &
+  LASSIE_PID=$!
+
+  exec node --max-old-space-size=2048 src/bin/shim.js &
+  SHIM_PID=$!
+
+  wait -n $LASSIE_PID $SHIM_PID
```
DiegoRBaquero, line 53: Following [docker doc](https://docs.docker.com/config/containers/multi-service_container/) 

```suggestion
  wait -n $LASSIE_PID $SHIM_PID
  exit $?
```

container/shim/src/fetchers/lassie.js
```diff
@@ -1,3 +1,143 @@
...
+
+  const lassieReq = _http
+    .get(
+      lassieUrl,
+      {
+        agent,
+        timeout: LASSIE_TIMEOUT,
+        signal: controller.signal,
+      },
+      async (fetchRes) => {
```
juliangruber, line 58: If this `async` function fails, it crashes the process vs just making the req/res cycle fail. This is because `http.get()` accepts a callback, and doesn't have `async` function support built in.

Options are basically using `fetch` (or another lib with `async` support), `utils.promisify` and/or `events.once()`.

juliangruber (changes requested):

Dockerfile
guanzo: great point, thanks

container/shim/src/fetchers/lassie.js
```diff
@@ -1,3 +1,143 @@
...
+  const carBlockIterator = await CarBlockIterator.fromIterable(streamIn);
+  let count = 0;
+
+  for await (const { cid, bytes } of carBlockIterator) {
+    if (!validateCarBlock(cid, bytes)) {
+      streamOut.status(502);
+      break;
+    }
+
+    // Is it safe to assume the first block is the requested one?
```
rvagg, line 134: Pretty safe for now but you have an easy opportunity here to check the cid against the requested one don't you? If you have the requested cid handy then pass it into here and check it (safest would be do a Buffer compare of the multihash bytes or up-convert both to cidv1 and string compare their base32 form). You could also do a check on the CarBlockReader's roots that there is just one and it matches what you asked for. Worth doing some sanity checking that things are working as expected. If this ever breaks, you should raise it with us (lassie). It's not an explicit contract but perhaps we should make it one.

container/shim/src/fetchers/lassie.js
```diff
@@ -1,3 +1,143 @@
...
+});
+
 export function respondFromLassie(req, res, { cid, format }) {
-  // todo
+  debug(`Fetch ${req.path} from Lassie`);
+
+  const isRawFormat = format === "raw";
+  const isInBlockCache = isRawFormat && !req.params.path && blockCache.has(cid);
+  if (isInBlockCache) {
+    const block = blockCache.get(cid);
```
rvagg, line 34: For these cache operations, get and put, do a `cid.toV1().toString()`. Since you only care about the bytes all that really matters is the multihash (you could even extract that and use that as the key) and you want to dedupe content where possible. It's not super-rare that you'll see the same content addressed by a v1 and a v0 CID but they're the same content.
Doing it by multihash is even better because you're just returning bytes out of this and it doesn't matter what the rest of the CID says in that case. There is a bunch of content stored in Filecoin now as "raw" that is really something else; most of our blockstores address by multihash now so it shouldn't matter from a source-of-block-data perspective.

container/shim/src/fetchers/lassie.js
guanzo: > Pretty safe for now but you have an easy opportunity here to check the cid against the requested one don't you?
ya for sure but I was held back by the possibility of a file path in the URL, in which case I don't know which cid to match against until Lassie returns CAR files that includes the parent CIDs.

I will add some sanity checks for when there's no path.

container/shim/src/fetchers/lassie.js
guanzo: Idk why the function is `async` in the first place, it doesn't await anything. Same for http.get call in `ipfs-gateway.js`. Must be leftover code that wasn't cleaned up. Also I know there's an eslint rule that prevents `async` without `await`, but I guess it's not enabled for this repo.

container/shim/src/fetchers/lassie.js
guanzo: If `cid.toV1().toString()` works fine then I'd prefer that for DX, since the cache requires string keys. I'd rather look at CID keys than stringified bytes.

container/shim/src/fetchers/lassie.js
rvagg: I haven't followed in detail, but I think we're signed up for only verifiable CARs so even in the case of a path the first block should be the one requested. At least I think that's a reasonable thing to build in to a requirement.

container/shim/src/fetchers/lassie.js
rvagg: Yep, that's fair enough and probably what I'd do for a first pass at this.

Just so you're aware though, you could also be doing a CID-style multibase encoding of the multihash and it would probably be as cheap, perhaps cheaper than doing the CID stringification and you can do it in a more compact string form too to save bytes in the LRU's key storage. You'd also then be properly disconnecting codec and CID version from the content being requested - if upstream wants bytes-for-hash then you just serve them back, regardless of the codec because that's not your problem.

Screenshot 2023-02-08 at 10 38 58 am

(`baseEncode` leaves off the multibase prefix since it's not helpful if they're all the same base, and adds a string concat operation; `encode` would put the `m` on there for base64 multibase).

guanzo: Ok you've convinced me now that I've extracted a code example from you 😛

container/shim/src/fetchers/lassie.js
hannahhoward: Lassie always returns parent CDs. I think you are good.

juliangruber (commented):
The fixed usage of `http.get` and `async` functions LGTM 🙏

container/shim/src/fetchers/lassie.js
```diff
@@ -1,3 +1,166 @@
...
+      clearTimeout(timeout);
+      debug.extend("error")(`Timeout from Lassie for ${cid}`);
+      lassieReq.destroy();
+      res.destroy();
+    });
+
+  req.on("close", () => {
+    clearTimeout(timeout);
+    if (!res.writableEnded) {
+      debug.extend("error")("Client aborted early, terminating gateway request");
```
DiegoRBaquero, line 105:
```suggestion
      debug.extend("error")("Client aborted early, terminating lassie request");
```

DiegoRBaquero (approved):

guanzo:
sry idk why my commit signing broke, will fix

PR #221, opened 2023-02-16
Lassie metrics

guanzo:
Record Lassie request metrics and submit logs to the log ingestor.

Logs are stored in memory and submitted on an interval. Losing logs due to a crash is tolerable.

### DIFF
--- a/container/shim/src/fetchers/lassie.js
+++ b/container/shim/src/fetchers/lassie.js
@@ -200,7 +200,8 @@ async function getRequestedBlockFromCar(streamIn, streamOut, requestedCidV1, pat
   }
 }
 
-async function queueMetricsReport() {
+async function queueMetricsReport(newMetrics) {
+  metrics.push(newMetrics)
   const date = lastMetricsReportDate;
   const canReport = !date || new Date() - date > METRICS_REPORT_INTERVAL;
   if (!canReport || !metrics.length) {
--- a/container/shim/src/fetchers/lassie.js
+++ b/container/shim/src/fetchers/lassie.js
@@ -30,9 +30,9 @@ const blockCache = new LRU({
 });
 const cidToCacheKey = (cidObj) => base64.baseEncode(cidObj.multihash.bytes);
 
-const metrics = [];
-const METRICS_REPORT_INTERVAL = 5_000;
+let metrics = [];
 let lastMetricsReportDate = null;
+const METRICS_REPORT_INTERVAL = 5_000;
 
 export async function respondFromLassie(req, res, { cidObj, format }) {
   debug(`Fetch ${req.path}`);
@@ -199,23 +199,24 @@ async function getRequestedBlockFromCar(streamIn, streamOut, requestedCidV1, pat
   }
 }
 
-async function queueMetricsReport(newMetrics) {
-  metrics.push(newMetrics)
+async function queueMetricsReport(newMetric) {
+  metrics.push(newMetric);
+
   const date = lastMetricsReportDate;
   const canReport = !date || new Date() - date > METRICS_REPORT_INTERVAL;
-  if (!canReport || !metrics.length) {
+  if (!canReport) {
     return;
   }
 
-  const logsToReport = metrics.splice(0);
   const chunkSize = 500;
   const promises = [];
 
-  for (let i = 0; i < logsToReport.length; i += chunkSize) {
-    const logs = logsToReport.slice(i, i + chunkSize);
+  for (let i = 0; i < metrics.length; i += chunkSize) {
+    const logs = metrics.slice(i, i + chunkSize);
     promises.push(submitLassieLogs(logs));
   }
 
+  metrics = [];
   lastMetricsReportDate = new Date();
 
   try {
--- a/container/shim/src/fetchers/lassie.js
+++ b/container/shim/src/fetchers/lassie.js
@@ -1,6 +1,6 @@
 import http from "node:http";
 import https from "node:https";
-import { Transform } from "node:stream"
+import { Transform } from "node:stream";
 
 import { CarBlockIterator } from "@ipld/car";
 import LRU from "lru-cache";
@@ -14,8 +14,8 @@ import { proxyResponseHeaders, toUtf8 } from "../utils/http.js";
 import { debug as Debug } from "../utils/logging.js";
 
 const debug = Debug.extend("lassie");
-const debugErr = debug.extend("error")
-debug.enabled = false // temporary until Lassie is stable.
+const debugErr = debug.extend("error");
+debug.enabled = false; // temporary until Lassie is stable.
 
 const LASSIE_TIMEOUT = 120_000;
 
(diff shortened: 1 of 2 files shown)


container/shim/src/fetchers/lassie.js
```diff
@@ -167,11 +191,36 @@ async function getRequestedBlockFromCar(streamIn, streamOut, requestedCidV1, pat
...
+      sendBlockResponse(streamOut, bytes, cidV1.toString());
     } else {
       const cacheKey = cidToCacheKey(cidV1);
       blockCache.set(cacheKey, bytes);
     }
     count++;
   }
 }
+
+async function queueMetricsReport() {
```
juliangruber, line 203:
```suggestion
async function queueMetricsReport(newMetrics) {
  metrics.push(newMetrics)
```

If we add the metrics pushing here, it's in the future less likely we forget to call `queueMetricsReport` after modifying the metrics, or that we call it too often.

container/shim/src/fetchers/lassie.js
```diff
@@ -111,12 +123,24 @@ export async function respondFromLassie(req, res, { cidObj, format }) {
...
+      startTime,
+      ttfbTime,
+      endTime,
+      numBytesDownloaded,
+      requestId,
+      url: req.protocol + "://" + req.get("host") + req.originalUrl,
+      httpStatusCode: lassieRes?.status ?? null,
+      requestErr,
+    });
+    queueMetricsReport();
```
juliangruber, line 142:
```suggestion
    queueMetricsReport({
      startTime,
      ttfbTime,
      endTime,
      numBytesDownloaded,
      requestId,
      url: req.protocol + "://" + req.get("host") + req.originalUrl,
      httpStatusCode: lassieRes?.status ?? null,
      requestErr,
    });
```

see below

container/shim/src/fetchers/lassie.js
```diff
@@ -167,11 +191,36 @@ async function getRequestedBlockFromCar(streamIn, streamOut, requestedCidV1, pat
...
 }
+
+async function queueMetricsReport() {
+  const date = lastMetricsReportDate;
+  const canReport = !date || new Date() - date > METRICS_REPORT_INTERVAL;
+  if (!canReport || !metrics.length) {
+    return;
+  }
+
+  const logsToReport = metrics.splice(0);
```
juliangruber, line 210:
```suggestion
  const logsToReport = [...metrics.splice];
```

To me this is more semantic

container/shim/src/fetchers/lassie.js
```diff
@@ -167,11 +191,36 @@ async function getRequestedBlockFromCar(streamIn, streamOut, requestedCidV1, pat
...
       const cacheKey = cidToCacheKey(cidV1);
       blockCache.set(cacheKey, bytes);
     }
     count++;
   }
 }
+
+async function queueMetricsReport() {
+  const date = lastMetricsReportDate;
+  const canReport = !date || new Date() - date > METRICS_REPORT_INTERVAL;
```
juliangruber, line 205: instead of this function, would it be simpler to have a separate interval, that just reports metrics every x ms?

juliangruber: Using something like this, it would also be race free, no matter how long the promises take:

```js
(async () => {
  while (true) {
    try {
      await submitMetrics()
    } catch (err) {
      debugErr(err)
    }
    await setTimeout(interval)
  }
})()
```

This has the downside that metrics aren't reported in a regular interval, at the upside of no race conditions and controlled concurrency. Wdyt?

juliangruber (changes requested):

container/shim/src/fetchers/lassie.js
DiegoRBaquero, line 210:
```suggestion
  const logsToReport = [...metrics];
```

container/shim/src/fetchers/lassie.js
```diff
@@ -167,11 +191,36 @@ async function getRequestedBlockFromCar(streamIn, streamOut, requestedCidV1, pat
...
+
+  for (let i = 0; i < logsToReport.length; i += chunkSize) {
+    const logs = logsToReport.slice(i, i + chunkSize);
+    promises.push(submitLassieLogs(logs));
+  }
+
+  lastMetricsReportDate = new Date();
+
+  try {
+    await Promise.all(promises);
```
DiegoRBaquero, line 222: Why batch them if they are all concurrent?

container/shim/src/fetchers/lassie.js
guanzo: Splice also empties the array, which needs to happen at some point.

container/shim/src/fetchers/lassie.js
guanzo: It is simpler, but I generally dislike having intervals floating in the global scope. If this module gets imported then the interval starts and could start spitting errors. 
`queueMetricsReport` only gets called when a request happens, which feels cleaner to me. 

Could go either way though. @DiegoRBaquero thoughts?

container/shim/src/fetchers/lassie.js
juliangruber: Gotcha. Since both Diego and me missed this, what do you think about making this more explicit instead?

container/shim/src/fetchers/lassie.js
guanzo: Considering the worst case scenario where the request exceeds the Lambda POST payload limit (6 MB)

container/shim/src/fetchers/lassie.js
DiegoRBaquero: Interval is a no no for me, I normally just make the functions create a timeout to themselves to account for that. Imho, for this which is not core (unlike nginx log ingesting and submission), I prefer the queuing when it happens, and as we flush it we can switch it to self-timeout as in the rest of the repo's "intervals"

container/shim/src/fetchers/lassie.js
DiegoRBaquero: ah, slice vs splice always gets me!

guanzo: How about this?
```js
  const logsToReport = metrics;
  metrics = []
  ```

guanzo: hopefully fixed by https://github.com/filecoin-saturn/L1-node/pull/221/commits/c90c0307a1c3cd05948377c2c8a67be7a578231a?

container/shim/src/fetchers/lassie.js
DiegoRBaquero: Yes, but one could mean multiple calls at the same time, the other is one "call" on an interval, if for whatever reason there's a race condition then it could execute multiple times. 

Right now it doesn't happen because it's all sync code until the setting of `lastMetricsReportDate`, but what if the `lastMetricsReportDate` was set after an `await` ever? BOOM

guanzo: >  but what if the lastMetricsReportDate was set after an await ever? BOOM

yep that's why it's set before the await. 

@juliangruber do we have a consensus here for the requested changes? It requires approval. I'm on a time limit today >.<

DiegoRBaquero (approved):

container/shim/src/fetchers/lassie.js
guanzo: sry need to bypass your approval.

juliangruber: all good please go ahead

PR #233, opened 2023-02-22
fix: use pipeline to pipe lassie response through metrics

rvagg:
pipeline() will wire up error events properly; specifically in this case an abort from node-fetch won't propagate through metricsStream and will be lost entirely, not even bubbling up as an error through respondFromLassie().

There's an error callback that could be used here, perhaps for logging? But for now at least the error will come back through the CarBlockIterator for-await and back up through the call stack.

### DIFF
--- a/container/shim/src/fetchers/lassie.js
+++ b/container/shim/src/fetchers/lassie.js
@@ -1,6 +1,6 @@
 import http from "node:http";
 import https from "node:https";
-import { Transform } from "node:stream";
+import { Transform, pipeline } from "node:stream";
 
 import { CarBlockIterator } from "@ipld/car";
 import LRU from "lru-cache";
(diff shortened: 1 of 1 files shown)


guanzo:
Thanks, completely forgot about error handling on the `pipe`, will review in the mornin'.

container/shim/src/fetchers/lassie.js
```diff
@@ -105,13 +105,13 @@ export async function respondFromLassie(req, res, { cidObj, format }) {
     } else {
       res.set("Cache-Control", "public, max-age=29030400, immutable");
     }
-    const carStream = lassieRes.body.pipe(metricsStream);
+    pipeline(lassieRes.body, metricsStream, (err) => {});
```
juliangruber, line 108: Just double checking, do we want to not log errors here?

juliangruber (approved):

container/shim/src/fetchers/lassie.js
guanzo: We do. That's something I can implement post merge

guanzo: After testing, errors are caught by the below `catch` block. An empty catch handler seems fine here?

rvagg: yeah, that's the point of the empty callback - those errors end up coming back through the async iterator from js-car anyway; but now you have multiple options! not confusing at all

guanzo: Yep i added one. Also worth noting that if `pipeline` throws an error if the last argument isn't a callback, so you have to pass one.

rvagg: fwiw you could also just grab the version from `'node:stream/promises'` and drop the returned Promise, no callback needed. This is more explicit I think though.

PR #238, opened 2023-02-23
Upgrade lassie v0.4.5

hannahhoward:
# Goals

Upgrades to a version of Lassie with better libp2p settings to reduce latency. 

# Implementation

Also introduces LASSIE_MAX_BLOCKS_PER_REQUEST = 10000. This number should be discussed and agreed upon. Default chunking for IPFS is usually 256K so this comes in around ~4GB, which seems generous for now.


guanzo:
4GB seems fine. Its easy to change if necessary.

PR #249, opened 2023-02-28
Add back L2 response path

juliangruber:
Without this, L2s in testnet receive no traffic.

However, Lassie will still be called also, since `L2_FIRE_AND_FORGET` is on

### DIFF
--- a/container/shim/src/index.js
+++ b/container/shim/src/index.js
@@ -72,12 +72,12 @@ const handleCID = asyncHandler(async (req, res) => {
   const isSampled = Math.random() < LASSIE_SAMPLE_RATE;
   const useLassie = isBifrostGateway || isSampled;
 
-  if (useLassie && LASSIE_ORIGIN) {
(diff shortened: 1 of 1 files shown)


DiegoRBaquero (approved):

container/shim/src/index.js
```diff
@@ -72,13 +72,13 @@ const handleCID = asyncHandler(async (req, res) => {
   const isSampled = Math.random() < LASSIE_SAMPLE_RATE;
   const useLassie = isBifrostGateway || isSampled;
 
-  if (useLassie && LASSIE_ORIGIN) {
-    return respondFromLassie(req, res, { cidObj, format });
-  }
-
   if (SATURN_NETWORK !== "main" && !req.params.path && (await maybeRespondFromL2(req, res, { cid, format }))) {
```
guanzo, line 75: can you move this block above line 71 so all the lassie logic is together?

Issue #250, opened 2023-03-01
L1 nodes should return a meaningful RetryAfter value when returning 429s


aarshkshah1992:
See https://github.com/filecoin-saturn/caboose/pull/59#discussion_r1120761107

guanzo:
done!

PR #252, opened 2023-03-01
send Retry-After header for 429 responses

guanzo:
Current rate limit is 100 req/sec with 200 burst. `Retry-After` value is set to 10 seconds. Thoughts?

Based on this SO post: https://serverfault.com/a/979156/657072

Handles https://github.com/filecoin-saturn/L1-node/issues/250


DiegoRBaquero (approved):

DiegoRBaquero:
Let's start with 10s as baseline and modify as needed

PR #257, opened 2023-03-03
Update lassie 0.6.1

hannahhoward:
# Goals

Updates lassie with NDJSON streaming(!) and brings us in compliance with the Graph Requests as CARs doc

Also includes temporary shutdown of db metrics

Deploy before weekend if possible


guanzo (approved):
lgtm

PR #259, opened 2023-03-04
Check OCSP status of cert

DiegoRBaquero:


container/shim/src/modules/registration.js
```diff
@@ -139,6 +140,30 @@ export async function register(initial = false) {
       }
     }
 
+    const certString = certBuffer.toString();
+    const cert = certString.substring(0, certString.indexOf("-----END CERTIFICATE-----") + 25);
+    const caCert = certString.substring(certString.indexOf("-----END CERTIFICATE-----") + 25);
+
+    if (caCert) {
+      try {
+        const response = await new Promise((resolve, reject) => {
```
guanzo, line 149: `promisify` is cleaner than a wrapper `Promise`

container/shim/src/modules/registration.js
```diff
@@ -139,6 +140,30 @@ export async function register(initial = false) {
       }
     }
 
+    const certString = certBuffer.toString();
+    const cert = certString.substring(0, certString.indexOf("-----END CERTIFICATE-----") + 25);
+    const caCert = certString.substring(certString.indexOf("-----END CERTIFICATE-----") + 25);
```
guanzo, line 145: Use variables to stay DRY and avoid magic number `25`

```suggestion
const boundary = "-----END CERTIFICATE-----"
const boundaryEndIndex = certString.indexOf(boundary) + boundary.length
const cert = certString.substring(0, boundaryEndIndex);
const caCert = certString.substring(boundaryEndIndex);
```

DiegoRBaquero: Is it really worth it? It's so specific to the string itself, not an external value, it's the length of ----END CERT...

guanzo: My thinking:

`certString.indexOf("-----END CERTIFICATE-----") + 25` is just pasted twice.. that should at least be put into a variable. It's an easy win for code clarity, there's no downside.

`const boundaryEndIndex = certString.indexOf("-----END CERTIFICATE-----") + 25`

Now, what is `25`? My first guess is its the length of `-----END CERTIFICATE-----`, but I had to spend a few seconds confirming it. Save everyone a few seconds by putting it into a descriptive variable name, so I end up with 

```js
const boundary = "-----END CERTIFICATE-----"
const boundaryEndIndex = certString.indexOf(boundary) + boundary.length
```

It might not be worth it for just these 2 lines alone, but this sort of "is it really worth it" mentality can spread over the entire code base and build up technical debt over time. It's better to just do it clean the first time. Make it readable for someone who's just skimming the code.

container/shim/src/modules/registration.js
```diff
@@ -141,19 +141,14 @@ export async function register(initial = false) {
...
       try {
-        const response = await new Promise((resolve, reject) => {
-          check(cert, caCert, (err, res) => {
-            if (err) {
-              reject(err);
-            }
-            resolve(res);
-          });
-        });
+        const response = await check(cert, caCert);
```
guanzo, line 151: arguments don't match the function parameters

container/shim/src/modules/tls.js
```diff
@@ -13,6 +13,7 @@ export const BACKUP_CERT_PATH = `${SSL_PATH}/node.backup.crt`;
 export const BACKUP_KEY_PATH = `${SSL_PATH}/node.backup.key`;
 
 export const certExists = await fsPromises.stat(CERT_PATH).catch((_) => false);
+export const backupCertExists = await fsPromises.stat(BACKUP_CERT_PATH).catch((_) => false);
```
guanzo, line 16: This isn't used anywhere, is that intended?

DiegoRBaquero: Yes, I had included it to conditionally do our logic, but I want the logs. Thing is, not everyone has a backup cert yet, and we can only swap certs once they have them, so this variable will be used.

guanzo (approved):
lgtm. run on testnet over the weekend and deploy on monday?

PR #260, opened 2023-03-04
Fix using Badbits

joaosa:
- This covers checking root cids;
- To do that we install [njs](https://nginx.org/en/docs/njs/) and uses `auth_request` to perform CID validation;
- Currently, it only matches root CIDs;
   - we need to add converting v0 to v1;
   - we need to support CID paths.
- Added very simple integration tests to cover a negative and positive CID test cases.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -26,7 +26,8 @@ jobs:
           cache-to: type=inline
           build-args: |
             SATURN_NETWORK=local
-            ORCHESTRATOR_URL=http://127.0.0.1
+            ORCHESTRATOR_URL=http://localhost
+            LOG_INGESTOR_URL=http://localhost
 
       - name: Run tests
         run: |
@@ -34,4 +35,10 @@ jobs:
 
           export HTTP_PORT=8080
           ORCHESTRATOR_REGISTRATION=false sh scripts/run.sh &
-          sh scripts/test_badbits.sh "$HTTP_PORT"
+
+          url="http://localhost:${HTTP_PORT}"
+          # wait for the container to be up
+          until curl -s "$url"; do
+              sleep 1;
+          done;
+          sh scripts/test_badbits.sh "$url"
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -3,11 +3,13 @@ name: Run integration tests
 on:
   push:
     branches: ["main"]
+    paths:
+      - ".github/workflows/integration-tests.yml"
+      - "container/nginx/**"
+      - "Dockerfile"
   pull_request:
     # The branches below must be a subset of the branches above
     branches: ["main"]
-  schedule:
-    - cron: "24 3 * * 1"
 
 jobs:
   badbits:
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -39,6 +39,6 @@ jobs:
           url="http://localhost:${HTTP_PORT}"
           # wait for the container to be up
           until curl -s "$url" -o /dev/null; do
-              sleep 1;
+              sleep 5;
           done
           sh scripts/test_badbits.sh "$url"
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,0 +1,50 @@
+name: Run integration tests
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: ["main"]
+  schedule:
+    - cron: "24 3 * * 1"
+
+jobs:
+  badbits:
+    name: Validate Badbits
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Build local Docker image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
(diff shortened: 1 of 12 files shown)


container/nginx/njs/badbits.js
```diff
@@ -0,0 +1,34 @@
+const fs = require("fs");
+const crypto = require("crypto");
+
+const badbits = {};
+const data = JSON.parse(fs.readFileSync("/etc/nginx/denylist.json"));
+data.forEach((entry) => (badbits[entry.anchor] = true));
```
juliangruber, line 6: Could also use Sets here:

```js
const badbits = new Set()
badbits.add(entry.anchor)
badbits.has(hashedCID)
```

container/nginx/njs/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "njs",
+  "type": "module",
```
juliangruber, line 3:
```suggestion
  "type": "module",
  "private": true,
  "license": "UNLICENSED"
```

to prevent accidental `npm publish`

container/nginx/njs/badbits.js
joaosa: That makes a lot of sense. Sadly, it will yield `2023/03/05 13:57:11 [error] 41#41: *1 js exception: ReferenceError: "Set" is not defined`. I had initially tried with a `Map`.

juliangruber: ah dang it's too old a version then :|

container/nginx/conf.d/shared.conf
```diff
@@ -24,12 +24,20 @@ location ~ ^/(ipns|api)/ {
     }
 }
 
+location /badbits {
+    internal;
+    set $cid_path $request_uri;
+    js_content main.filterCID;
+}
```
juliangruber, line 31: it would be great to add a test for this route

container/nginx/njs/badbits.js
```diff
@@ -0,0 +1,34 @@
+const fs = require("fs");
+const crypto = require("crypto");
```
juliangruber, line 2:
```suggestion
import fs from "node:fs";
import crypto from "node:crypto";
```

I'm not sure why it worked before, but one shouldn't mix commonjs and esm module syntax

container/nginx/njs/badbits.js
```diff
@@ -0,0 +1,34 @@
...
+const badbits = {};
+const data = JSON.parse(fs.readFileSync("/etc/nginx/denylist.json"));
+data.forEach((entry) => (badbits[entry.anchor] = true));
+
+const ipfsRegex = /^\/ipfs\/(\w+)(\/?.*)/;
+
+// TODO implement matching CID paths
+// TODO convert CID v0 to CID v1
+// implementation ref: https://github.com/protocol/bifrost-infra/blob/af46340bd830728b38a0ea632ca517d04277f78c/ansible/roles/nginx_conf_denylist/files/lua/helpers.lua#L80
+function filterCID(req) {
```
juliangruber, line 13:
```suggestion
export function filterCID(req) {
```

container/nginx/njs/badbits.js
```diff
@@ -0,0 +1,34 @@
...
+    .digest("hex");
+
+  if (badbits[hashedCID]) {
+    req.return(403);
+  } else {
+    req.return(200);
+  }
+}
+
+export default { filterCID };
```
juliangruber, line 34:
```suggestion
```

juliangruber (changes requested):

container/nginx/njs/badbits.js
joaosa: njs isn't exactly node.js, but its own thins, so scoping by `node:*` won't work

container/nginx/njs/badbits.js
joaosa: see below

container/nginx/njs/badbits.js
joaosa: njs only supports `export default {foo, bar}` syntax as per [here](https://github.com/nginx/njs/issues/121#issuecomment-968085559).

container/nginx/njs/badbits.js
joaosa: ended up going with `export default function ...`

juliangruber (changes requested):
looks great to me, just needs tests

DiegoRBaquero (approved):

juliangruber (approved):
awesome work!!

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,44 @@
+name: Run integration tests
+
+on:
+  push:
+    branches: ["main"]
```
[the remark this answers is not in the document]
joaosa: It seems to be the default for this repo. Since this is so important it might make sense to run it as often as possible 🤔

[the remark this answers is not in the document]
DiegoRBaquero: it should run on any (main push or pull request) IF Dockerfile or nginx folder are modified, wdyt?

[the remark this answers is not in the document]
joaosa: addressed this. it doesn't seem like this is possible for `pull_request` events

DiegoRBaquero (approved):

guanzo (approved):
lgtm. great work

PR #262, opened 2023-03-06
Run Lassie publicly

joaosa:
Run on 0.0.0.0, so we can scrape metrics.


DiegoRBaquero:
Operators are not required to have the port open. Just a note.

hannahhoward:
@DiegoRBaquero I think metric scraping is just trusted nodes anyway no?

hannahhoward:
Also though, I wonder if we should throw the metrics only on a separate port to avoid exposing the whole api?

DiegoRBaquero:
> @DiegoRBaquero I think metric scraping is just trusted nodes anyway no?

Ah ok, that should be fine then

DiegoRBaquero:
> Also though, I wonder if we should throw the metrics only on a separate port to avoid exposing the whole api?

This would allow anyone to (ab)use L1s' lassie to fetch content or DDoS them. I agree we should only expose metrics

joaosa:
> > Also though, I wonder if we should throw the metrics only on a separate port to avoid exposing the whole api?
> 
> This would allow anyone to (ab)use L1s' lassie to fetch content or DDoS them. I agree we should only expose metrics

That is a good point. We should only expose metrics if possible.

joaosa:
Closing this, as we don't need it anymore

PR #264, opened 2023-03-06
Dynamic cache min uses depending on disk usage

DiegoRBaquero:
If disk usage is less than 90%, allow caching on first hit


joaosa (approved):
LGTM

PR #266, opened 2023-03-06
Swap certs if revoked

DiegoRBaquero:


guanzo:
Orchestrator isn't aware of cert revocations so it's going to issue revoked cert to all new nodes, right?

https://github.com/filecoin-saturn/L1-node/pull/266/files#diff-b7d97ee156eb86cc2d8e2ffdbb1abc72faadf7bacd4524d9a2ba1bcf7997b66cL104

DiegoRBaquero:
> Orchestrator isn't aware of cert revocations so it's going to issue revoked cert to all new nodes, right?
> 
> https://github.com/filecoin-saturn/L1-node/pull/266/files#diff-b7d97ee156eb86cc2d8e2ffdbb1abc72faadf7bacd4524d9a2ba1bcf7997b66cL104
> 
> When the new node receives certs, it'll restart, then check if certs are revoked. If so, it'll swap and restart again. So it doesn't successfully register until it's using the unrevoked certs. Does that sound right?

That would be correct, if both certs are revoked, then it'll crash loop, because it can't serve traffic anyways.

guanzo (approved):
lgtm

PR #273, opened 2023-03-09
Cleanup system stats

DiegoRBaquero:
Disk stats are now MB only and memory KB only.

### DIFF
--- a/container/shim/src/utils/system.js
+++ b/container/shim/src/utils/system.js
@@ -10,7 +10,7 @@ const exec = promisify(CpExec);
 
 const debug = Debug.extend("system");
 
-const meminfoKBToGB = (bytes) => bytes / 1_000_000;
+const meminfoKBToGB = (bytes) => (bytes / 1_000_000).toFixed(0);
 
 export async function getMemoryStats() {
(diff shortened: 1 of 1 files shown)


container/shim/src/utils/system.js
```diff
@@ -10,7 +10,7 @@ const exec = promisify(CpExec);
 
 const debug = Debug.extend("system");
 
-const meminfoKBToGB = (bytes) => bytes / 1_000_000;
+const meminfoKBToGB = (bytes) => (bytes / 1_000_000).toFixed(0);
```
guanzo, line 13: Use powers of 2 instead of 10? e.g. `1048576` (1024 ** 2) instead of `1_000_000`

DiegoRBaquero: Nope, reports incorrect info. Base 10 reports correctly for both memory and disk

DiegoRBaquero: 4027844 kB -> 4GB using base 10, 3.84 GB using base 2.

guanzo (approved):
lgtm

PR #277, opened 2023-03-10
Reply with 403 for manually blocked content

joaosa:
- So it's conformant with Badbits;
- Added a test to cover this;
- Further [ref](https://filecoinproject.slack.com/archives/C031K3BG527/p1678433424047959?thread_ts=1678232596.144739&cid=C031K3BG527)

p.s. disregard the PR branch name 😅

### DIFF
--- a/container/nginx/conf.d/shared.conf
+++ b/container/nginx/conf.d/shared.conf
@@ -62,7 +62,7 @@ location / {
     # $arg_filename, $arg_download, $param_go_get, $http_if_none_match
     proxy_cache_key             $uri$format_final$arg_depth$arg_bytes$arg_filename$arg_download$param_go_get$http_if_none_match;
     proxy_cache_lock            on;
-    proxy_cache_valid           200 301 302 410     365d;
+    proxy_cache_valid           200 301 302 365d;
     proxy_cache_valid           any                 1m;
     proxy_cache_min_uses        $cache_min_uses;
     proxy_cache_lock_age        30m;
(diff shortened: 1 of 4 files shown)


container/nginx/conf.d/shared.conf
```diff
@@ -62,7 +62,7 @@ location / {
     # $arg_filename, $arg_download, $param_go_get, $http_if_none_match
     proxy_cache_key             $uri$format_final$arg_depth$arg_bytes$arg_filename$arg_download$param_go_get$http_if_none_match;
     proxy_cache_lock            on;
-    proxy_cache_valid           200 301 302 410     365d;
```
DiegoRBaquero, line 65: Shouldn't we cache 403s then?

joaosa: I supposed we could, but won't that interfere with other 403s on the shim codebase, such as [here](https://github.com/filecoin-saturn/L1-node/blob/main/container/shim/src/modules/registration.js#L234)? - I think this is the only place.

DiegoRBaquero: Has it's own proxying config https://github.com/filecoin-saturn/L1-node/blob/51582eeba5b5453f1900978339ccf5289590c524/container/nginx/conf.d/register.conf

joaosa: addressed

DiegoRBaquero (approved):

Issue #225, opened 2023-02-16
Remove SATURN_NETWORK environment variable


DiegoRBaquero:
From docker compose and .env template

It should only come from the arg in the build.

Could it stop being a docker env variable and pass it from the arg to a file maybe?

gruns:
> From docker compose and .env template

and the docs 🙂 (it's pervasive throughout the l1 readme and docs)

joaosa:
Leaving this in the docs, as it's the way to specify if operators are fetching a test net or main net version. The issue is preventing this from getting passed into the container which https://github.com/filecoin-saturn/L1-node/pull/248 takes care of.

joaosa:
This is no longer present as it can be seen [here](https://raw.githubusercontent.com/filecoin-saturn/L1-node/main/docker-compose.yml). Also deployed this change to all core nodes

PR #281, opened 2023-03-15
Load feature-flags for staged deployments

joaosa:
- Make use of https://docs.docker.com/compose/compose-file/#env_file to load any env vars into the container;
- This will unfortunately load `SATURN_NETWORK` as well, but I figured this isn't a big issue given we need to work on https://github.com/filecoin-saturn/L1-node/issues/225#issuecomment-1460805305.


guanzo:
What about users that don't use a `.env` file to set env vars?

joaosa:
> What about users that don't use a `.env` file to set env vars? For ex. they set them in `~/.profile`, will those still get loaded into the container?

Using the `.env` file is the way to go now, but yes that is a valid point. We can leave the other config behind just in case 👍

guanzo:
Yeah we need to keep our changes backwards compatible. 

> I addressed this.

I don't see any differences from your original commit >.<

joaosa:
> Yeah we need to keep our changes backwards compatible.
> 
> > I addressed this.
> 
> I don't see any differences from your original commit >.<

My bad 🤦 . It should be fine now

guanzo (approved):

PR #283, opened 2023-03-16
Fix log ingestor

DiegoRBaquero:
- Add log rotation
- Fix glob
- Improve offset setting no valid logs
- keep alive agent


container/logrotate/nginx
```diff
@@ -0,0 +1,15 @@
...
+        rotate 30
+        compress
+        delaycompress
+        notifempty
+        create 640 nginx nginx
+        sharedscripts
+        postrotate
+                if [ -f /var/run/nginx.pid ]; then
+                        kill -USR1 `cat /var/run/nginx.pid`
+                fi
```
guanzo, line 13: Is this killing the nginx process?

joaosa: It should re-open nginx log files. more [here](http://nginx.org/en/docs/control.html)

guanzo:
lgtm but I don't know how well the current log ingestor handles log rotations. I'd like to run this on canary for a bit

guanzo (approved):

PR #287, opened 2023-03-20
Disable connections on Elastic IPFS for now

hannahhoward:
Updates lassie and adds to the excluded list the main peer for Elastic IPFS, which we need to disable cause of the bill we're running up :)


gruns (approved):
lgtm 👍

PR #288, opened 2023-03-20
Lower block limit, default depth=1

hannahhoward:
# Goals

Reduce Lassie over-fetching

# Implementation

- set a block limit of 10 for now
- use only shallow recursion by default for now


container/shim/src/fetchers/lassie.js
```diff
@@ -69,10 +69,17 @@ export async function respondFromLassie(req, res, { cidObj, format }) {
...
+    if (key === "depth" && val === "all") {
+      newKey = "depthType";
+      newVal = "full";
+    }
     lassieUrl.searchParams.set(newKey, toUtf8(newVal));
   }
   lassieUrl.searchParams.set("format", "car");
-
+  // if no depth type set, default to shallow
+  if (!lassieUrl.searchParams.get("depthType")) {
```
guanzo, line 80:
```suggestion
  if (!lassieUrl.searchParams.has("depthType")) {
```

guanzo (approved):
LGTM.

Something to consider: 

"you CANNOT request more than 10 blocks period. even with depth=all."

PR #291, opened 2023-03-22
Custom speedtest server ID

DVerdeV:
In the command used to measure the speed (speedtest) it is necessary to add "--server-id" to take into account the server ID. Otherwise, it will ignore it and select it automatically.

In docker-compose.yml it is added, but personally I have ignored it and that has led me to confusion. I think it would be better to indicate only the ID in the docker-compose.yml.


DiegoRBaquero:
Should go inside the env var

PR #309, opened 2023-03-29
Add server-timing header, nginx and shim

DiegoRBaquero:
Adds `Server-Timing` header for nginx times and `Server-Timing` header for shim

headers:
image

as seen on network console:
image


container/shim/src/fetchers/lassie.js
```diff
@@ -38,6 +38,8 @@ const METRICS_REPORT_INTERVAL = 5_000;
 export async function respondFromLassie(req, res, { cidObj, format }) {
   debug(`Fetch ${req.path}`);
 
+  res.startTime("shim/lassie");
```
DiegoRBaquero, line 41: Started, ends when requests ends

container/shim/src/fetchers/lassie.js
```diff
@@ -91,7 +93,9 @@ export async function respondFromLassie(req, res, { cidObj, format }) {
         "X-Request-ID": requestId,
       },
     };
+    res.startTime("shim_lassie_headers");
```
DiegoRBaquero, line 96: This is connect + wait for headers time

guanzo (approved):

PR #319, opened 2023-04-03
logs fix

### LINKED ISSUE #317: Readline.js bug due log logrorate
Hi, this commit add logrorate https://github.com/filecoin-saturn/L1-node/commit/a2d19081490ac03b3e27ebd0a71c19a7ad561ec3. But I find my some nodes would not submit retrievals entire hour. I have opened a issue but you ingored it. So I read the code of readline.js [https://github.com/filecoin-saturn/L1-node/blob/main/container/shim/src/utils/readlines.js](url).
It use hashmap of log file's inode. However if a new node-access.log file happens to have the same inode as an old log file. Readline.js will always read zero line becase .offsetfile's offset number is the size of an old log file. 
So it will submit no retrieval util the new node-access.log' size goes bigger than that old log file

DiegoRBaquero:
- Last 24 hours of logs
- Reset offset if stat size is smaller than cached

Closes #317

### DIFF
--- a/container/logrotate/nginx
+++ b/container/logrotate/nginx
@@ -1,7 +1,7 @@
 /usr/src/app/shared/nginx_log/*.log {
         hourly
         missingok
-        rotate 30
+        rotate 24
         compress
         delaycompress
         notifempty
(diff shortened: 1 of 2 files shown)


guanzo:
We can also make the mapping key in `.offsetfile` to be (inode + file creation time) to ensure uniqueness.

guanzo (approved):

DiegoRBaquero:
That transition will need to be in two steps as to not lose any logs. For now let's go with this

PR #324, opened 2023-04-07
update lassie v0.8.1

hannahhoward:
# Goals

Update Lassie to support car-scope parameter, a few other things

# For discussion

The conversion of depth parameter to car-scope in lassie.js is cause I think ARC may still be sending only depth=1

### DIFF
--- a/container/shim/src/fetchers/lassie.js
+++ b/container/shim/src/fetchers/lassie.js
@@ -160,13 +160,17 @@ function createLassieURL(req, isRawFormat) {
     // translate depth parameter for lassie
     let newKey = key;
     let newVal = val;
-    if (key === "depth" && (val === "1" || val === "0")) {
-      newKey = "depthType";
-      newVal = "shallow";
+    if (key === "depth" && val === "1") {
+      newKey = "car-scope";
+      newVal = "file";
+    }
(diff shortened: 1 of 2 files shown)


guanzo (approved):

container/shim/src/fetchers/lassie.js
```diff
@@ -160,13 +160,17 @@ function createLassieURL(req, isRawFormat) {
...
-    if (key === "depth" && (val === "1" || val === "0")) {
-      newKey = "depthType";
-      newVal = "shallow";
+    if (key === "depth" && val === "1") {
+      newKey = "car-scope";
+      newVal = "file";
+    }
+    if (key === "depth" && val === "0") {
+      newKey = "car-scope";
+      newVal = "root";
```
lidel, line 169: @hannahhoward is this a typo or did we implement different things? 

https://github.com/ipfs/bifrost-gateway/issues/80 → https://github.com/ipfs/bifrost-gateway/pull/81/files → I did `block` instead of `root` as we used both durign discussion, and `block` felt more descriptive  :see_no_evil:   

(not the end of the world, no need to revert releease, bifrost-gateway still sends  `depth=0`  so it will work, but we should clean this up before we remove `depth`)

PR #331, opened 2023-04-10
Fix the watchtower pre-check random wait

joaosa:
I noticed this was broken when revisiting this for allowing faster updates on testnet. To fix this: 

- I added running the l1 container with `--init` so signals can be properly caught as we're running multiple processes;
- I also added a new env var to configure the order of magnitude of the random wait;
- I used a script for the random wait as that is the supported way to do this with watchtower.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -128,7 +128,8 @@ COPY container/shim/package*.json ./
 RUN npm ci --production --ignore-scripts
 
 # copy the generated modules and all other files to the container
-COPY container/start.sh ./
+COPY --chmod=0744 container/start.sh ./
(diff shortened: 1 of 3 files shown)


guanzo:
huh, so this is why deployments were rolling out so quickly

guanzo (approved):

joaosa:
@guanzo @DiegoRBaquero We cannot merge this or it'll break auto-update. This is because docker-compose.yml changes make it to the mainnet as soon as they're merged. I'm going to try to find a way to make this a bit saner.

DiegoRBaquero:
Let's merge this tomorrow or monday

PR #336, opened 2023-04-11
make LASSIE_EXCLUDE_PROVIDERS easier to override

guanzo:
Make it a build arg so it can be overridden in one operation (set build arg) instead of two (set build arg && override env var in start.sh)


tchardin (approved):
Thanks!

PR #354, opened 2023-05-04
update lassie to v0.9.0

kylehuntsman:
### Updates
Aggregates event metrics and fixes minor env var and log level bugs
https://github.com/filecoin-project/lassie/releases/tag/v0.9.0

### ! Additional Work Required !
This PR also needs to update the event recorder URL secret to use the new aggregate endpoint. It's the same endpoint, just using `/v2/...` instead of `/v1/...`. I wasn't sure where to go to update that secret.

### Deployment Strategy
We'd like this to be deployed to Canary nodes for a significant amount of time before being pushed up to production so that we have ample time to collect enough metrics to determine that everything seems normal. Ideally the metrics don't change from their current trends. I'll be around until 2023-05-04 04:00 UTC today to check up on things if this get's deployed soon, otherwise I'll be online tomorrow at 2023-05-04 15:00 UTC as well. We're looking to get this out this week before we release HTTP on Lassie.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -131,7 +131,7 @@ RUN curl -fsSL https://deb.nodesource.com/setup_18.x | bash - \
 
 # Download lassie
 ARG TARGETPLATFORM
-ARG LASSIE_VERSION="v0.8.2"
+ARG LASSIE_VERSION="v0.9.0"
 RUN if [ "$TARGETPLATFORM" = "linux/amd64" ]; then ARCHITECTURE=amd64; \
   elif [ "$TARGETPLATFORM" = "linux/arm64" ]; then ARCHITECTURE=arm64; \
   else ARCHITECTURE=386; fi \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -167,9 +167,6 @@ ARG ORCHESTRATOR_URL
 ARG LASSIE_EVENT_RECORDER_AUTH
 ARG LASSIE_EVENT_RECORDER_URL
 
-# Use random peerId until this is fixed https://github.com/filecoin-project/lassie/issues/191
-ARG LASSIE_EXCLUDE_PROVIDERS="QmcCtpf7ERQWyvDT8RMYWCMjzE74b7HscB3F8gDp5d5yS6"
```
guanzo, line 171: I think we still need the variable declaration `ARG LASSIE_EXCLUDE_PROVIDERS`, because it's referenced below. We can still remove the default value though.

DiegoRBaquero: Is this no longer used? Or what is the reason to remove it?

kylehuntsman: Than can we just also remove L186? I'm fine with keeping this. I was following the comment on L170. The bug mentioned in the comment was fixed, so I was going to remove it.

guanzo (approved):

DiegoRBaquero (approved):

PR #365, opened 2023-05-08
removes block verification as lassie already does it

guanzo:
Also resolves the issue where blake3 hashed blocks would fail to verify, as L1 didn't support blake3.


joaosa (approved):
lgtm

PR #381, opened 2023-05-15
feat(deps): update lassie to v0.10.0

willscott:


Dockerfile
```diff
@@ -182,8 +182,6 @@ ENV VERSION_HASH=$VERSION_HASH
 ENV ORCHESTRATOR_URL=$ORCHESTRATOR_URL
 
 ENV LASSIE_EVENT_RECORDER_AUTH=$LASSIE_EVENT_RECORDER_AUTH
-ENV LASSIE_EVENT_RECORDER_URL=$LASSIE_EVENT_RECORDER_URL
```
DiegoRBaquero, line 185: Should we also remove the ARGs in lines 167 and 168?

PR #382, opened 2023-05-15
Revert "renice lassie"

DiegoRBaquero:
Reverts filecoin-saturn/L1-node#372

### DIFF
--- a/container/start.sh
+++ b/container/start.sh
@@ -79,7 +79,6 @@ if [ "${LASSIE_ORIGIN:-}" != "" ]; then
     lassie daemon &
   fi
   LASSIE_PID=$!
-  renice +10 $LASSIE_PID
 
   node --max-old-space-size=4096 /usr/src/app/src/bin/shim.js &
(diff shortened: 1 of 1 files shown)


juliangruber:
why was this reverted?

DiegoRBaquero:
Given that Lassie is single process, we want to allow Lassie to consume as much CPU as it can without being re-scheduled by its nice priority. It was giving funky CPU distribution values.

Before: `%Cpu(s):  6.8 us,  6.3 sy, 82.6 ni,  1.4 id,  0.1 wa,  0.0 hi,  2.9 si,  0.0 st`
After: `%Cpu(s): 86.5 us,  6.6 sy,  0.0 ni,  3.8 id,  0.0 wa,  0.0 hi,  3.1 si,  0.0 st`

It also spiked timeouts, because all the cache miss will have a hard time discovering peers and fetching content as the process is choked, under CPU constraints.

juliangruber:
Got it, thank you!

PR #397, opened 2023-05-22
Write an unclean end to car files on error

willscott:
cc @hannahhoward  - i think this is the js equivalent for https://github.com/filecoin-project/lassie/issues/245

### DIFF
--- a/container/shim/src/fetchers/lassie.js
+++ b/container/shim/src/fetchers/lassie.js
@@ -127,11 +127,21 @@ export async function respondFromLassie(req, res, { cidObj, format }) {
     if (controller.signal.aborted) {
       debugErr(`Timeout for ${cid} after ${new Date() - startTime}ms`);
 
-      if (!res.headersSent) res.sendStatus(504);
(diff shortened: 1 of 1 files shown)


guanzo:
I can confirm that nginx will not cache the response with "unclean end", however the error is not propagated to the client. Likely because nginx ensures a clean end is written.

guanzo (approved):

DiegoRBaquero:
During my local tests, the client (browser) did get an error when connection was destroyed without proper close.

guanzo:
My testing was flawed: the browser client does receive errors if the nginx upstream writes an unclean end.

over http2, the chrome error is `net::ERR_HTTP2_PROTOCOL_ERROR 200`
over https1.1 and plaintext http, the chrome error is `net::ERR_INCOMPLETE_CHUNKED_ENCODING 200 (OK)`

A `fetch` call will complete without error, but if you try to read the body it'll throw `TypeError: Failed to fetch`

image

PR #418, opened 2023-06-02
Add a base L1 bifrost docker compose

joaosa:
Add a basic docker compose file so the bifrost and L1 setup is reproducible locally.

1. Clone the https://github.com/ipfs/bifrost-gateway into `../../../ipfs/bifrost-gateway` (relative to this project)

2. Make sure to generate the L1 certs for localhost with:

```
cd $HOME/shared/ssl
mkcert -key-file node.key -cert-file node.crt localhost
```

3. `cd integration && docker compose up`

4. Fire requests to bifrost (e.g.)

```
curl -v 127.0.0.1:8081/ipfs/$CID -r 0-1048575 # range request
curl -v 127.0.0.1:8081/ipfs/$CID # regular block request
```

### DIFF
--- a/integration/docker-compose.yml
+++ b/integration/docker-compose.yml
@@ -30,5 +30,5 @@ services:
       STRN_ORCHESTRATOR_URL: https://orchestrator.strn.pl
       CABOOSE_BACKEND_OVERRIDE: saturn-node
     ports:
-      - 8083:8081
-      - 8042:8041
+      - 8081:8081
+      - 8041:8041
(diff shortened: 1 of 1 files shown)


aarshkshah1992:
@joaosa Why do we need those L1 certs on localhost ?

joaosa:
> @joaosa Why do we need those L1 certs on localhost ?

Because Caboose attempts to use the L1's HTTPS endpoint. If we can configure it to use HTTP then this gets simpler

guanzo (approved):

DiegoRBaquero (approved):

PR #429, opened 2023-06-09
New range request version with correct support for end bound ranges

tchardin:
Please note that this version will not work as expected for ranges with a start offset. The goal is to get a first version filtering end bound ranges and seeing if we see an improvement with the most common 1Mb range requests.


tchardin:
@guanzo I see no real issue with response not ending cleanly, is there a test we could add in there to double check? Otherwise integration tests are looking fine to me for a release.

guanzo (approved):

PR #440, opened 2023-06-15
Remove concurrent retrievals

hannahhoward:
# Goals

The current version of onion is likely getting rate limited on requests to DAG House. The was removed in the main branch a couple days ago. But looks like it got copied over in the onion tests. Seems like it might make sense to figure out how to not have these env vars in three places -- maybe a multi-stage build or something?


joaosa:
Thanks @hannahhoward ! yeah that makes a lot of sense. I'll cleanup things a bit. I guess a multi-stage build should be the way

joaosa (approved):
LGTM

PR #457, opened 2023-07-14
Clean up old CAR files at 10% chance

DiegoRBaquero:


container/start.sh
```diff
@@ -25,6 +25,13 @@ echo "$(date -u) [container] Disk: $(df -h /usr/src/app/shared | awk '(NR>1)')"
 mkdir -p /usr/src/app/shared/ssl
 mkdir -p /usr/src/app/shared/nginx_log
 
+# Clean up old cached CAR files with 10% chance and if not already cleaned up
+if [ "$RANDOM" -lt 3277 && ! -f "/usr/src/app/shared/cleaned_cache" ]; then
```
willscott, line 29: where does random come from?

willscott (approved):

container/start.sh
DiegoRBaquero: > $RANDOM is an internal Bash [function](https://tldp.org/LDP/abs/html/functions.html#FUNCTIONREF) (not a constant) that returns a pseudorandom [[1]](https://tldp.org/LDP/abs/html/randomvar.html#FTN.AEN5817) integer in the range 0 - 32767.

PR #464, opened 2023-08-07
Add the onion integration config

joaosa:
Just adds the boilerplate to the main repo


guanzo (approved):

guanzo:
actually, i kind of want to move `shim.Dockerfile` and `shim_start.sh` to be inside the `onion` dir because it's only used in that context and i'd rather not have them appear in the root dir

container/shim/src/bin/shim.js
```diff
@@ -55,7 +55,7 @@ setTimeout(async function () {
 }, 500);
 
 const server = http.createServer(app);
-server.listen(PORT, "127.0.0.1", async () => {
+server.listen(PORT, "0.0.0.0", async () => {
```
joaosa, line 58: Unsure if this change is ok to merge. any concerns?

joaosa:
closing this as it no longer matters

PR #481, opened 2023-08-28
docs(faq): Add initial details about penalties

vorburger:
Ad discussed on Slack, and related to https://github.com/31z4/saturn-moonlet/issues/3.

@DiegoRBaquero @AmeanAsad

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -60,6 +60,21 @@ These are the current penalties that affect both DNS weight and earned FIL:
 - Fraudulent logging (e.g. self-dealing)
 - Multi-noding (Running multiple nodes on the same host)
 
+These are visibile e.g. by hovering over the _Weight_ of a Node on https://dashboard.saturn.tech, or on the _Penalty_ graph of the [Moonlet](https://github.com/31z4/saturn-moonlet).
+
+Here are more details about each kind of penalty, with information what could cause it, and how to remedy if your Node encounters them:
+
+- `error_ratio` is caused by errors as shown in the log of L1 Node container, scuh as:
+  - Node refusing to connect with a client multiple times.
+- `dup_cache_miss_ratio` is caused if a computed error ratio for consecutive duplicate cache misses goes over a threshold.
+- `health_check_failures` are caused by an unexpectedly (not deregistered) unreachable node.
+  - This penalty starts with each health check failure event, and then gradually decreases over ~6h.
+  - Fix this by resolving the root cause of the node unavailability.
+  - Note the [deregister my node](#how-can-i-manually-deregister-my-node) section.
+
+None of these are "expected" under normal operation, and all are something that you want to keep down.
+
+
 ## Registration
 
 ### My Node fails to register with error ETIMEDOUT/EHOSTUNREACH


docs/faq.md
```diff
@@ -60,6 +60,21 @@ These are the current penalties that affect both DNS weight and earned FIL:
 - Fraudulent logging (e.g. self-dealing)
 - Multi-noding (Running multiple nodes on the same host)
 
+These are visibile e.g. by hovering over the _Weight_ of a Node on https://dashboard.saturn.tech, or on the _Penalty_ graph of the [Moonlet](https://github.com/31z4/saturn-moonlet).
```
DiegoRBaquero, line 63: Given Moonlet is external, it should not be mentioned in the official L1 FAQ

docs/faq.md
```diff
@@ -60,6 +60,21 @@ These are the current penalties that affect both DNS weight and earned FIL:
...
+
+Here are more details about each kind of penalty, with information what could cause it, and how to remedy if your Node encounters them:
+
+- `error_ratio` is caused by errors as shown in the log of L1 Node container, scuh as:
```
DiegoRBaquero, line 67: `errorRatio1h` and `errorRatio12h` are the actual values

```suggestion
- Error ratio is caused by network errors connecting to the node and HTTP 5xx errors
```

docs/faq.md
```diff
@@ -60,6 +60,21 @@ These are the current penalties that affect both DNS weight and earned FIL:
...
+  - Node refusing to connect with a client multiple times.
+- `dup_cache_miss_ratio` is caused if a computed error ratio for consecutive duplicate cache misses goes over a threshold.
```
DiegoRBaquero, line 69:
```suggestion
- Duplicate Cache Miss Ratio is caused when the same CID & file path is cached missed to upstream providers. This shouldn't happen if the node has enough disk and the right file permissions.
```

docs/faq.md
```diff
@@ -60,6 +60,21 @@ These are the current penalties that affect both DNS weight and earned FIL:
...
+- `health_check_failures` are caused by an unexpectedly (not deregistered) unreachable node.
```
DiegoRBaquero, line 70:
```suggestion
- Health Check Failures are caused by an unexpectedly (not deregistered) unreachable node. This isn't necessarily the node's fault, but could also be a connectivity issue anywhere in the internet path from the orchestrator to the node.
```

docs/faq.md
```diff
@@ -60,6 +60,21 @@ These are the current penalties that affect both DNS weight and earned FIL:
...
+  - This penalty starts with each health check failure event, and then gradually decreases over ~6h.
+  - Fix this by resolving the root cause of the node unavailability.
```
DiegoRBaquero, line 72: This can be anywhere between Orchestrator (USA), submarine cables, country, DC, router, node. Not always easy to spot

docs/faq.md
```diff
@@ -60,6 +60,21 @@ These are the current penalties that affect both DNS weight and earned FIL:
...
+  - Note the [deregister my node](#how-can-i-manually-deregister-my-node) section.
+
+None of these are "expected" under normal operation, and all are something that you want to keep down.
```
DiegoRBaquero, line 75: Error rates are expected. Duplicate cache misses are not, this statement isn't general

vorburger: Oh! OK, I have removed this phrase, but... if Error rates are expected under normal operations, then why does Node get penalized for it, if there is nothing its operator can do about it?

vorburger:
@DiegoRBaquero thanks for the prompt review; I've incorporated your feedback - merge this?

PR #498, opened 2023-09-28
feat: remove car range module, passthrough to newer lassie

DiegoRBaquero:
Removes CAR range module
Updates Lassie to latest version supporting CAR ranges


DiegoRBaquero:
No, this PR deactivates it :)

PR #499, opened 2023-10-04
doc(requirements): require unique /24 block per geolocation

DiegoRBaquero:
Starting October, each node must have a unique /24 block in its geolocation. This is meant to prevent same operators from running multiple nodes in the same geolocation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ on [Filecoin Slack](https://filecoinproject.slack.com/)! 👋
 
 ### Node hardware requirements
 
-- Linux server with a static public IPv4 address
+- Linux server with a static public IPv4 address in a unique /24 CIDR block
 - Root access / passwordless sudo user ([How to](https://askubuntu.com/questions/147241/execute-sudo-without-password))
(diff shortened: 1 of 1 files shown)


guanzo (approved):

README.md
```diff
@@ -42,7 +42,7 @@ on [Filecoin Slack](https://filecoinproject.slack.com/)! 👋
 
 ### Node hardware requirements
 
-- Linux server with a static public IPv4 address
+- Linux server with a static public IPv4 address in a unique /24 CIDR block
```
guanzo, line 45: Will the typical node op understand "/24 CIDR block"?

DiegoRBaquero: God I hope so, but easily google-able, also open to suggestions

31z4: Can you elaborate on the uniqueness? Is it unique across the whole Saturn network? Or unique for a specific operator's wallet address?

AmeanAsad: @DiegoRBaquero maybe best to just provide an example here to supplement the docs on the requirement.

PR #504, opened 2023-10-10
Renew cert with .1% chance

DiegoRBaquero:
Start renewing certs for per-node direct access

### DIFF
--- a/container/shim/src/modules/registration.js
+++ b/container/shim/src/modules/registration.js
@@ -193,6 +193,16 @@ async function checkCertValidity(certBuffer, registerOptions) {
     valid = false;
   }
 
+  if (
+    NETWORK === "main" &&
+    cert.subjectAltName &&
+    !cert.subjectAltName.includes(".l1s.saturn.ms") &&
+    Math.random() < 1 / 1000
+  ) {
+    debug("Certificate is missing <ip>.l1s.saturn.ms SAN, getting a new one...");
+    valid = false;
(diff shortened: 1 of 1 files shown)


container/shim/src/modules/registration.js
```diff
@@ -193,6 +193,16 @@ async function checkCertValidity(certBuffer, registerOptions) {
     valid = false;
   }
 
+  if (
+    NETWORK === "main" &&
+    cert.subjectAltName &&
+    !cert.subjectAltName.includes(".l1s.saturn.ms") &&
```
hannahhoward, line 199: comment here would be nice to describe intent.

If I read right, this is:
- you have a cert and your cert is not unique to your node (i.e. new naming scheming .l1s.saturn.ms)
- your roll a 1 on a D1000 dice :)
- then we mark the cert invalid and get a new one.

And this runs every 8 min on every node? So that means for 3000 nodes we're doing about 3 nodes every 8 minutes -- is that our target?

hannahhoward (commented):
Can you confirm I've understood intent correctly before merging?

container/shim/src/modules/registration.js
DiegoRBaquero: Yes for most. The validity check is only on initial registration or every 24 hours, so 3 nodes every update / day

hannahhoward (approved):
LGTM

PR #507, opened 2023-10-13
Check for unique SAN in cert, dial down to 1% chance every 12 hours

DiegoRBaquero:
Check more often, with less chance and that the specific SAN (if initial) doesn't exist

### DIFF
--- a/container/shim/src/modules/registration.js
+++ b/container/shim/src/modules/registration.js
@@ -98,9 +98,9 @@ export async function register(initial = false) {
 
   const certBuffer = await fsPromises.readFile(CERT_PATH);
 
-  // Check cert validity on initial registration and at least once daily
-  if (initial || lastInitialRegistration < Date.now() - 24 * 60 * 60 * 1000) {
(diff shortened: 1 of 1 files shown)


container/shim/src/modules/registration.js
```diff
@@ -98,9 +98,9 @@ export async function register(initial = false) {
 
   const certBuffer = await fsPromises.readFile(CERT_PATH);
 
-  // Check cert validity on initial registration and at least once daily
-  if (initial || lastInitialRegistration < Date.now() - 24 * 60 * 60 * 1000) {
-    await checkCertValidity(certBuffer, registerOptions);
+  // Check cert validity on initial registration and at least twice daily
```
AmeanAsad, line 101: not to be nit picky here but we already have consts in the orch defined for times so might as well use them here.

AmeanAsad: ohh thought this was the orch repo, my bad

AmeanAsad (approved):

guanzo (approved):

Issue #494, opened 2023-09-25
[METERING] Extract JWT in each request and attach to L1 ingestor log submission


tchardin:
were going to break this into two milestones

### milestone 1

* [x] implement authentication based on session token JWTs such that each request gets logged appropriately under the correct customer's account for metering, but not billing, purposes

### milestone 2

* [ ] l1s validate session token JWTs and either accept or reject requests without having to make an external network request 
* [x] arc and gateway (caboose) traffic will need to be authenticated, too 👏 

why break the above into two milestones?
* if we launch at lab week without l1s validating session JWTs, effectively all bandwidth is free and there's no reason for customers to pay us or select a plan beyond the free plan. however, this is acceptable and preferable to not launching at all if we don't have time to also implement milestone 2
* ideally, we launch with both milestone 1 and milestone 2, where with milestone 2 customers will have to pay once they hit their plan's cap (or pay us as part of the pay-as-you-go plan)

guanzo:
> arc and gateway (caboose) traffic will need to be authenticated, too

Does this mean that all requests will become authenticated in the future? So you can't simply `curl` an L1 anymore?

guanzo:
JWTs are being extracted from nginx access logs and submitted to the log-ingestor which stores it into postgres.

wilkyr31d:
@guanzo I'm super assuming this can be closed now, but if I'm wrong re-open.

PR #509, opened 2023-10-20
Verify JWT

guanzo:
Handles https://github.com/filecoin-saturn/L1-node/issues/494

Verifies JWTs with https://github.com/max-lt/nginx-jwt-module

JWTs can be submitted as a query param or in the `Authorization` header.

Requests will fail if:
* JWT is missing/expired/invalid.
* `Origin` header is missing
* `Origin` is not allowed by JWT `allow_list`

⚠This PR cannot be deployed to production until the prod JWT lambda is running and arc clients are updated to use prod JWTs.


container/nginx/njs/auth.js
```diff
@@ -0,0 +1,73 @@
...
+
+  /* eslint-disable-next-line no-undef */
+  return hashedCID in denylist;
+}
+
+function isAllowedDomain(req) {
+  const allowListStr = req.variables.jwt_claim_allow_list;
+  // Only browser requests are allowed for now.
+  const requestOrigin = req.variables.http_origin;
+  if (!allowListStr || !requestOrigin) {
```
guanzo, line 41: Requests without an `Origin` header will fail. Is this what we want?

Browsers will automatically send an `Origin` header in most cases. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin. For Saturn, all requests in the browser should be cross origin fetches, so the `Origin` header should be added.

container/nginx/jwt_pub.key
```diff
@@ -0,0 +1,4 @@
+-----BEGIN PUBLIC KEY-----
```
guanzo, line 1: This is the staging public key.

DiegoRBaquero (approved):

Dockerfile
```diff
@@ -64,6 +81,20 @@ RUN echo "Cloning njs $NJS_VERSION" \
   && ./configure \
   && make
 
+RUN echo "Cloning nginx-jwt-module $NGINX_JWT_VERSION" \
+  && git clone --depth 1 --branch $NGINX_JWT_VERSION https://github.com/max-lt/nginx-jwt-module.git
+
+RUN echo "Installing libjwt $LIBJWT_VERSION" \
```
hannahhoward, line 87: oof libjwt is a compile from source? Have we checked how long this takes? I do see though that 1.15.3 doesn't appear to be available on Debian stable, so maybe this is just what we have to do.

container/nginx/jwt_pub.key
hannahhoward: that's reasonable. ultimately, it's a public key so security shouldn't need to be that tight.

container/nginx/njs/auth.js
hannahhoward: kind of nuetral on this -- what does it get us? I would generally expect [*] to allow empty origin, but not particularly strong on this.

hannahhoward: one worry -- what about non-browser requests? we're already working on native clients for games. they can of course still set origins.

scripts/integration_tests.sh
```diff
@@ -42,3 +55,36 @@ test_range_request "QmafUYju2Ab4ETi5HJG1cqjmnjs2xw9PUuBKzU7Hi3zvXU/MC_TheSource.
 
 # range request with offset
 test_range_request "bafybeifpz6onienrgwvb3mw5rg7piq5jh63ystjn7s5wk6ttezy2gy5xwu/Mexico.JPG?entity-bytes=1048576:2097152"
+
+################
+# JWT Auth
+################
+
+authentication_err=401 # jwt missing or invalid
```
hannahhoward, line 63: ❤️ on all these tests

hannahhoward (approved):
all of my comments are not blocking.

container/nginx/njs/auth.js
guanzo: > kind of nuetral on this -- what does it get us? 

I checked with @tchardin and this seems to be the current requirement, but I'll admit it feels weird. 

> I would generally expect [*] to allow empty origin, but not particularly strong on this.

I had the same thought, I can add this.

> one worry -- what about non-browser requests? we're already working on native clients for games. they can of course still set origins.

Origin header checking is involved in both JWT creation and verification, it's very browser centric atm. The non-browser auth flow is still TBD I guess.

gruns: * the current flow for lab week is for browser clients only. we're purposefully not onboarding any customers beyond browsers for the private beta launch

* we 100% need to support non-browser clients in the future, and making every client, browser or not, set an `Origin` header is one way to do it, but my gut strongly says not the best way. its hacky and non-obvious to overload `Origin`

* in my mind, the right/better approach is to update the JWT in the future when it's handed out to non-browser clients -- those that don't send an `Origin` header. then the l1 could:
  * first check the JWT to see what kind of JWT it is. is it one for a website with an allowlist of domains? if so, check for the `Origin` header and test for a match 👍 
  * if the JWT doesn't have an allowlist of domains, its for an app. like a native game. and the auth check will be different 
  
* we can change which claims are included in the JWT's claims payloads and they can differ depending on what kind of client is making the request

* for native apps, perhaps instead of an `Origin` header, we could use a different, custom header, like `Saturn-App-Id`, that gets set by the smart client in the native app, and this `Saturn-App-Id` is corroborated with the custom JWT claim we add for native apps. but, as of now, this hasn't been designed

either way, i dont think we need to solve this problem in the future once we get a better idea of non-browser use cases. fundamentally, the entire point of having auth is so we can track, meter, and bill customers for their usage. and, where possible, prevent abuse by say game B using game A's saturn account to distribute content. the latter becomes harder when every client can 'lie' easily, which isn't possible at scale across a whole website with the `Origin` header. this requires some more thinking and exploring how saturn will be used by non-browser clients

gruns:
🎉 👏

PR #520, opened 2024-01-16
Do both platforms in parallel, update actions versions

DiegoRBaquero:
Pending increase of concurrency for larger worker.


joaosa (approved):
🙏

PR #521, opened 2024-01-17
node

DiegoRBaquero:
- Upgrade node
- Upgrade deps


AmeanAsad:
@DiegoRBaquero a few of these dep upgrades are major version upgrades. Just wanted to double check that our test cases are good enough to catch any regressions due to breaking changed in those upgrades?

DiegoRBaquero:
@AmeanAsad They are not good enough, but I did went and checked each major release notes to check there's nothing breaking

joaosa (approved):
🙏

PR #527, opened 2024-02-01
fix: improve error handling on shim

DiegoRBaquero:


container/shim/src/bin/shim.js
```diff
@@ -43,7 +47,7 @@ setTimeout(async function () {
     await register(true).catch((err) => {
       debug(`Failed to register ${err.name} ${err.message}`);
       // we don't try again if we fail the initial registration
```
joaosa, line 49: Do we want to try again now?

joaosa (commented):
Is this to reduce the reported errors on Sentry?

container/shim/src/bin/shim.js
DiegoRBaquero: Exit code was changed because we currently have `restart: unless-stopped` so I prefer the correct exit code.

DiegoRBaquero:
> Is this to reduce the reported errors on Sentry?

Yes, because Sentry captures unhandled exceptions and rejections, and because this is the entry point, these were considered unhandled

joaosa (approved):

